PR #13, opened 2015-02-03
Feature/Platform-550 User registration

ArturKlajnerok:
User registration process that cointans:
- User account validation and creation
- Sending confirmation email by wikia api.
- Antispoof support by wikia api.
- Phalanx username and email blocking.

ping: @pchojnacki @jcellary @Grunny


helios/register_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	user, err := models.NewUser(r.Form.Get("username"), r.Form.Get("password"), r.Form.Get("email"), r.Form.Get("birthdate"))
+
+	if err != nil {
+		resp.Output["error"] = E_VALIDATION_ERROR
+		resp.Output["error_description"] = err.Error()
+		return
+	}
+
+	if blocked := controller.checkPhalanxBlock(resp, user.Name); blocked {
+		return
```
pchojnacki: Shouldn't we at least try to return some error message ?

helios/register_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	if spoofed := controller.checkAntiSpoof(resp, user.Name); spoofed {
+		return
```
pchojnacki: Same here

helios/register_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	if !controller.disableUserAccountCreation {
+		controller.createAccount(resp, user)
+	} else {
+		resp.Output["success"] = true
+		resp.Output["success_description"] = "User valid"
+	}
+}
+
+func (controller *RegisterController) createAccount(resp *osin.Response, user *models.User) {
+	user.Name = utils.FirstToUpper(user.Name)
```
pchojnacki: I understand that this is MW specific transformation.
Maybe adding function that will be self explanatory like. func(...) transformUserMwStyle(user* models.User) { user.Name = utils.FirstToUpper(user.Name) }
that would make this transformation easy to spot and reason about

helios/register_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	err := controller.userStorage.CreateUser(user)
+
+	if err != nil {
+		controller.setServerError(resp, err)
+	} else {
+		if send, err := controller.wikiaClient.SendConfirmationEmail(user.Name); !send {
```
pchojnacki: I think if we are not depending on the output of those functions. then they should be done on a separate thread than user registration. So that they don't block the registration process.

models/user.go
```diff
@@ -3,34 +3,167 @@ package models
...
+	DefaultOptions         = ""
+	HashPrefix             = ":B:"
+	MinAge                 = 13
+	MinLen                 = 1
+	MaxLen                 = 50
+	SaltSize               = 8
 )
 
+var (
+	ErrUserAlreadyExist  = errors.New("User already exist")
```
pchojnacki: What about translations of those strings?

helios/register_controller.go
ArturKlajnerok: errors are set in checkPhalanxBlock body

helios/register_controller.go
ArturKlajnerok: errors are set in checkAntiSpoof body

pchojnacki: Ah ok. Then its ok.

utils/utils.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func IsIP(s string) bool {
+	return net.ParseIP(s) != nil
+}
+
+func RandomHex(size int) string {
+	b := make([]byte, size/2)
+	_, err := rand.Read(b)
+	if err != nil {
```
jcellary: Minor: you can write if _, err := rand.Read(b); err != nil {

utils/utils.go
```diff
@@ -0,0 +1,74 @@
...
+		logger.GetLogger().ErrorErr(err)
+	}
+
+	return hex.EncodeToString(b)
+}
+
+func YearsPassed(since, to time.Time) int {
+	years := 0
+
+	for since.AddDate(years, 0, 0).Before(to) {
```
jcellary: This is an O(n) loop for calculating years passed, whereas you can easly deduce number of years using arithmetic operations which will have O(1)

models/user.go
```diff
@@ -3,34 +3,167 @@ package models
...
+
+	err = user.validateBirthDate()
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (user *User) validateName() error {
```
jcellary: You start some functions with upper case, some with lower case. You should start all with upper to be consistent.

helios/register_controller.go
```diff
@@ -0,0 +1,191 @@
+package helios
```
jcellary: You did not provide any e2e (integration) test for this controller. I think it would be good to have one. Of course one difficulty is writing to db and calling other services, but you already have that to some extend for shadow mode, so maybe it wouldn't be so difficult?

helios/register_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	timer := utils.CreateTimerForAPICall(controller.influxdbClient, AppName, "registerHandler")
+	defer utils.CloseTimer(timer)
+
+	resp := controller.server.NewResponse()
+	defer resp.Close()
+	defer osin.OutputJSON(resp, w, r)
+
+	err := r.ParseForm()
+	if err != nil {
+		resp.Output["error"] = osin.E_INVALID_REQUEST
```
jcellary: You defined a function setValidationError (which btw. should start with a capital letter), it should be used consistently in all places where you set output parameters. Same few lines below.

helios/register_controller.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	if spoofed := controller.checkAntiSpoof(resp, user.Name); spoofed {
+		return
+	}
+
+	if !controller.disableUserAccountCreation {
+		controller.createAccount(resp, user)
+	} else {
+		resp.Output["success"] = true
```
jcellary: I would recommend defining a function which sets these parameters similarly to errors to have a single place where you set all output parameters to have better visibility into what are the possible fields of the response message.

helios/register_controller.go
```diff
@@ -0,0 +1,191 @@
...
+		resp.Output["error"] = osin.E_INVALID_REQUEST
+		resp.Output["error_description"] = controller.deferror.Get(osin.E_INVALID_REQUEST)
+		return
+	}
+
+	if access := controller.checkClientAccess(resp, r.Form.Get("client_id"), r.Form.Get("client_secret")); !access {
+		return
+	}
+
+	user, err := models.NewUser(r.Form.Get("username"), r.Form.Get("password"), r.Form.Get("email"), r.Form.Get("birthdate"))
```
jcellary: This function is pretty long, maybe putting the first two blocks into separate functions like you did with phalanx and anti spoof checks would make it more readable and simpler?

models/user.go
```diff
@@ -3,34 +3,167 @@ package models
...
+
+func (user *User) validateBirthDate() error {
+	if utils.YearsPassed(user.BirthDate, time.Now()) < MinAge {
+		return ErrBirthDateMinAge
+	}
+
+	return nil
 }
 
 func (user *User) IsValidPassword(password string) bool {
```
jcellary: You should not be using log.Printf, but our kibana log instead

models/user.go
```diff
@@ -48,6 +181,15 @@ func (user *User) IsValidPassword(password string) bool {
...
+	return t.Format(RegistrationTimeFormat)
+}
+
+func saltedPasswordHash(password, salt string) string {
+	return HashPrefix + salt + ":" + HashPassword(password, salt)
+}
+
 func ExtractHashAndSalt(hash string) (string, string) {
 	splitedHash := strings.SplitN(hash[3:], ":", 2)
 	if len(splitedHash) != 2 {
```
jcellary: You should not be using log.Printf, but our kibana log instead

wikia/wikia.go
```diff
@@ -0,0 +1,93 @@
...
+	err := client.callWithToken(endpoint, data, response)
+	if err != nil {
+		return false, err
+	}
+
+	return response.Success, nil
+}
+
+func (client *Client) SendConfirmationEmail(name string) (bool, error) {
+	endpoint := "wikia.php?controller=Wikia\\Helios\\HelperController&method=sendConfirmationEmail"
```
pchojnacki: If you want you can directly smtp to sendgrid to send this.

ArturKlajnerok: I know, but there are some templates, translations and logic connected with confirmation email in our MW extension that we didn't wanted to move for now.

utils/utils.go
ArturKlajnerok: Considering that we need to opera on "time.Time" and consider variable days number in a given years this is a best solution I came up with using "time" package api. If you have more specific recommendation we can introduce it.

models/user.go
ArturKlajnerok: done

models/user.go
ArturKlajnerok: done

models/user.go
ArturKlajnerok: We need to solve internationalization use case for services in general. This is not a part of this feature.

helios/register_controller.go
ArturKlajnerok: Can you specify more precisely what should be extracted to separate functions. Note that this code was changed, so check the latest diff.

helios/register_controller.go
ArturKlajnerok: Tests included with usage of testify suite.

jcellary:
LGTM, one minor thing is that you could also provide tests for registration controller for the unhappy path (i.e. validation errorr - blocked by phalanx, user already existis, invalid email provided etc.)

ArturKlajnerok:
Tests for invalid requests where work in progress and are done now.

models/user.go
pchojnacki: ok

helios/register_controller.go
pchojnacki: ack

pchojnacki:
LGTM

PR #16, opened 2015-03-24
Platform-1078 Fix for user null fields parsing

ArturKlajnerok:
- Fix for possible birthdate null
- Fix for possible registration time null

Take a look: @jcellary @pchojnacki


jcellary:
Made minor fix, rest is ok, merging

PR #17, opened 2015-04-16
[SERVICES-209] Documentation

ArturKlajnerok:
Added basic documentation in readme and Api documentation for swagger.
Also small fixes and CORS support.
Take a look: @jcellary @pchojnacki


README.md
```diff
@@ -1,17 +1,46 @@
...
+
+## Building ##
+Build script is using dependencies stored with godep
+```bash
+./build
+```
 
 ## Running ##
-Pass mysql config as param:
+To run the service a valid `config.ini` need to be provided.
```
jcellary: needs

README.md
```diff
@@ -1,17 +1,46 @@
...
 * github.com/stretchr/testify      - Testing extension
 
+# API Docs
+Created with [swagger generator](https://github.com/yvasiyarov/swagger) from annotations.
+[Docs](https://github.com/Wikia/helios/blob/master/docs/docs.go) are availble with running instance of [Swagger](http://swagger.io/) documentation
+
+# Owners
+Currently owned by the services team
+
+## Installing ##
```
jcellary: I think the link to wikia one article abour deploying would be useful here

jcellary:
LGTM apart from minors. As for the docs I guess we'll know if they're comprehensive enough when someone tries to use the authorization API (unless there will be more information on that in the proxy documentation). I think there's still no full list of possible error codes that the service can return, but maybe I missed it.

ArturKlajnerok:
@jcellary Possible response list with error codes is on wikia one and include changes that will be introduced with http status codes

pchojnacki:
LGTM, nice work!

PR #18, opened 2015-04-20
[SERVICES-226] Http status codes

ArturKlajnerok:
Http status codes for all resources and tests for them. Also changed Redis storage to not return internal errors when specific key doesn't exist.

Take a look: @pchojnacki @jcellary @michalroszka


helios/oauth_controller.go
```diff
@@ -53,6 +53,12 @@ func (controller *OAuthController) infoHandler(w http.ResponseWriter, r *http.Re
 	if ir := controller.server.HandleInfoRequest(resp, r); ir != nil {
 		controller.server.FinishInfoRequest(resp, r, ir)
 		resp.Output["user_id"] = ir.AccessData.UserData
+	} else {
+		if resp.InternalError != nil {
+			resp.StatusCode = http.StatusInternalServerError
+		} else {
```
aquilax: You don't need else here :no_bell:

nandy-andy: Hey mate! :wave: Bored in Sweden, heh? :japanese_ogre:

aquilax: Ha ha a little. But my point is still valid :palm_tree:

ArturKlajnerok: Hey @aquilax ;), if handling info request fails and there's no internal error my default http status is "http.StatusBadRequest", so this else fits for that conditions :snail:

aquilax: @ArturKlajnerok the way I see it `resp.InternalError` is either nil or not so there are only two options for `resp.StatusCode` in this block `http.StatusInternalServerError` | `http.StatusBadRequest`

So this should be equivalent:
```go
resp.StatusCode = http.StatusBadRequest
if resp.InternalError != nil {
  resp.StatusCode = http.StatusInternalServerError
}
```

Cheers! :dizzy_face:

ArturKlajnerok: Yeah that make sense :panda_face:

jcellary:
LGTM

pchojnacki:
LGTM

PR #20, opened 2015-04-29
[SERVICES-297] Fixing tests setup

ArturKlajnerok:
Tests can now run with local dependencies stored in Godep. I've also added missing dependencies.

Take a look : @jcellary @pchojnacki @sqreek


test
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
jcellary: How does this file relate to run_tests.sh? It would seem they both try to acomplish the same thing, so I think it would be good to merge them. Also we should establish a convention - I was adding the .sh extension to bash script, this file doesn't have one, I think we should choose one option and be consistent.

sqreek:
LGTM

test
ArturKlajnerok: I will merge run_tests.sh to test script in new pull request as we need this one now to run the tests in CI. As for conventions I would vote for no ".sh" convention.

PR #21, opened 2015-05-04
Go all in on godeps & README updates

drsnyder:
This PR evolved as I learned about how go packages work and about the dependency issues in the go ecosystem. The fundamental change with this PR is to go all-in with our usage of `godep`.

Previously we were using the directory structure created by `godep` and the dependency definition while managing the `GOPATH` manually. IMO we should go all-in and rely on `godep` for managing the dependencies and the path.

This PR does that while also resolving the external dependency on `godep` and build, compile and test time. This should make the workflow more consistent with the `go` tooling so it "just works" without any settings from the dev or environment-- you just use `build` or `test` directly and the rest is handled for you transparently.

This has been tested in both dev and via the deploy tools. I was able to successfully start from scratch (with no `godep` installed) and deploy to dev.

/cc @Wikia/services-team


README.md
```diff
@@ -18,9 +18,14 @@ OAuth2 server compatible with mediawiki user accounts. Provides user authenticat
 * github.com/coopernurse/gorp      - ORM-ish library
 * github.com/stretchr/testify      - Testing extension
 
+## Development Dependencies
+* [godep](https://github.com/tools/godep)
```
ArturKlajnerok: I would add godep informations in "dependencies" section and describe that it should only use for maintenance/updates of dependencies and not for build and test.

README.md
```diff
@@ -43,10 +48,15 @@ To run the service a valid `config.ini` needs to be provided.
...
 
 ## Testing ##
-E2E test need a running instance of helios.
+
+All of the tests can be run through the test driver. To run the end to end
+tests you will need a running instance of helios.
+
 ```bash
-./test
+./run_tests.sh [all|unit|e2e]
```
ArturKlajnerok: We need to merge run_tests.sh to test and use the second one.

README.md
drsnyder: But this contradicts what's coded in the build script and in `run_tests.sh`. They both either reference paths generated by `godep` or `godep` directly.

README.md
drsnyder: Personally, I think it's better to just use the tools directly. For example:

```
godep go test -v github.com/Wikia/helios/helios
godep go test -v github.com/Wikia/helios/utils -run TestContainsBlockedChars
```

I updated all of the scripts to use `godep` directly in 4657ec11.

drsnyder:
For the [jenkis build](http://qa-s2.wikia-prod:8080/job/helios_tests/configure) to work properly we'll need to add `go get github.com/tools/godep` to the build script.

drsnyder:
Last call

ArturKlajnerok:
LGTM :+1:

PR #22, opened 2015-05-04
Add a cookie to the token response

drsnyder:
In this PR, I add a cookie to the token response and add functional tests for the cookie. This is to satisfy [SERVICES-290](https://wikia-inc.atlassian.net/browse/SERVICES-290).

/cc @Wikia/services-team 

TODO:
 * [x] Parse the expires date and confirm that it's :older_man:


helios/oauth_controller_test.go
```diff
@@ -0,0 +1,15 @@
+package helios
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+  "time"
```
aquilax: Formatting is off here :european_post_office:

drsnyder: ty :). fixed in b207aff

helios/oauth_controller.go
```diff
@@ -10,6 +10,11 @@ import (
 	"github.com/Wikia/helios/utils"
 	"github.com/influxdb/influxdb/client"
 	"net/http"
+	"time"
+)
+
+const (
+	AccessToken = "access_token"
```
ArturKlajnerok: We could name it AccessTokenKey to be more descriptive.

helios/oauth_controller.go
```diff
@@ -155,3 +162,17 @@ func (controller *OAuthController) setOutputError(resp *osin.Response, errorCode
...
+
+func addAccessTokenToCookie(resp *osin.Response, responseWriter http.ResponseWriter) {
+	access_token, exists := resp.Output["access_token"]
+	if exists {
+		cookie := createAccessTokenCookie(access_token.(string))
+		http.SetCookie(responseWriter, cookie)
+	}
+}
+
+
```
ArturKlajnerok: Redundant line :panda_face:

helios/oauth_controller.go
```diff
@@ -155,3 +162,17 @@ func (controller *OAuthController) setOutputError(resp *osin.Response, errorCode
...
+func createAccessTokenCookie(access_token string) (*http.Cookie) {
```
ArturKlajnerok: use camel case here access_token -> accessToken :camel:

ArturKlajnerok: I don't see any advantage here from returning the pointer. We could return regular struct : ```return http.Cookie{Name: AccessToken, Value: access_token, Expires: cookieExpireDate}``` and then use pointer like: ```http.SetCookie(responseWriter, &cookie)```. :cookie:

drsnyder: Both of these are addressed in 6a59b58.

ArturKlajnerok:
With small cosmetics changes LGTM :+1:

helios/oauth_controller.go
drsnyder: Fixed in 535b4e5

helios/oauth_controller.go
drsnyder: Ok, changed the return value in 4ac886f.

helios/oauth_controller.go
```diff
@@ -155,3 +162,16 @@ func (controller *OAuthController) setOutputError(resp *osin.Response, errorCode
...
+	accessToken, exists := resp.Output["access_token"]
+	if exists {
+		cookie := createAccessTokenCookie(accessToken.(string))
+		http.SetCookie(responseWriter, &cookie)
+	}
+}
+
+func createAccessTokenCookie(accessToken string) (http.Cookie) {
+	cookieExpireDate := time.Now().AddDate(10, 0, 0) // 10 years
+	return http.Cookie{Name: AccessTokenKey, Value: accessToken, Expires: cookieExpireDate}
```
Grunny: @drsnyder Since this is effectively a session cookie, can we set HttpOnly to true on it?

drsnyder: Good point. My general sense is that the cookie is incomplete. For example, we also aren't specifying a domain even though it should be implicit. I'll create a separate ticket for cleaning it up.

PR #25, opened 2015-05-06
Support for user new password validation

ArturKlajnerok:
This will fix problems with reset password functionality.

/cc @Wikia/services-team


models/user.go
```diff
@@ -181,6 +181,13 @@ func (user *User) IsValidPassword(password string) bool {
 		return false
 	}
 
+	if len(user.NewPassword) > 3 {
+		salt, passHash := ExtractHashAndSalt(user.NewPassword)
+		if passHash == HashPassword(password, salt) {
```
frankfarmer: This comparison is vulnerable to timing attacks, right?

http://verboselogging.com/2012/08/20/a-timing-attack-in-action

models/user.go
```diff
@@ -181,6 +181,13 @@ func (user *User) IsValidPassword(password string) bool {
...
+			return true
```
nmonterroso: should this return `false` if the extracted hash doesn't match the hash of the user input?

models/user.go
drsnyder: Below (not shown) the return value is `passHash == HashPassword(password, salt)`. For consistency, should the same be done here?

ArturKlajnerok: Thanks @frankfarmer I will look into that for general authentication security :panda_face:

models/user.go
ArturKlajnerok: @drsnyder & @nmonterroso The reason is to check if password matches `user_newpassword` and if not proceed and make an match with `user_password`.

nmonterroso: ah ok, so even after changing your password you can login with your old password

nmonterroso:
LGTM

drsnyder:
From the description, it's not clear the problem you are trying to solve. Can you say more about what the problem was? It's also not clear from the function what constitutes a valid password. 

I think there should also be some unit tests for the function and the added check. Without tests, future us won't know what we can or can't change.

ArturKlajnerok:
@drsnyder I've tried address all your feedback in : https://github.com/Wikia/helios/pull/26

PR #26, opened 2015-05-07
[SERVICES-311] Fix for reset password support

ArturKlajnerok:
This is covering changes and tests to user password hashing to meet this requirements:
- Latest reset password hash support
- Old reset password hash support
- Latest regular password hash support
- Old regular password hash support

User need to always have the hash for regular password but it's possible that hash for the reset password is not set.

/cc @Wikia/services-team


models/user.go
```diff
@@ -176,25 +177,35 @@ func (user *User) MediaWikiTransformation() {
...
+}
 
-	if len(user.NewPassword) > 3 {
-		salt, passHash := ExtractHashAndSalt(user.NewPassword)
-		if passHash == HashPassword(password, salt) {
-			return true
-		}
+func (user *User) passwordMatchesHash() bool {
+	if len(user.PasswordHashed) < MinHashSize {
+		logger.GetLogger().Error("To short password hash for user: " + user.Name)
```
aquilax: Looks like there is a typo here `To short...` => `Too short...`

models/user.go
```diff
@@ -176,25 +177,35 @@ func (user *User) MediaWikiTransformation() {
...
 	}
+	return user.passwordMatchesHash()
+}
 
-	if len(user.NewPassword) > 3 {
-		salt, passHash := ExtractHashAndSalt(user.NewPassword)
-		if passHash == HashPassword(password, salt) {
-			return true
-		}
+func (user *User) passwordMatchesHash() bool {
```
aquilax: Passing `user.Passsword` and `user.PasswordHashed` as parameters will make the funtcion easier to test. :telescope:

aquilax:
On a side note you may consider using http://golang.org/pkg/net/mail/#ParseAddressList for validating the email address :ghost:

models/user.go
```diff
@@ -176,25 +178,38 @@ func (user *User) MediaWikiTransformation() {
 }
 
 func (user *User) IsValidPassword(password string) bool {
-	if len(user.HashedPassword) < 3 {
-		logger.GetLogger().Error("To short hash for user: " + user.Name)
-		return false
+	user.Password = password
```
drsnyder: This looks odd. Why do you have to set the users password at the top of the function? I would not expect this function to have side effects or alter the state of the `user` struct.

models/user.go
```diff
@@ -176,25 +178,38 @@ func (user *User) MediaWikiTransformation() {
...
+
+	if user.passwordMatchesHash(user.ResetPasswordHashed) {
+		return true
 	}
+	return user.passwordMatchesHash(user.PasswordHashed)
```
drsnyder: These 4 lines could be simplified to `return user.passwordMatchesHash(user.ResetPasswordHashed) || user.passwordMatchesHash(user.PasswordHashed)`.

ArturKlajnerok: Changed in: f71a009

models/user.go
ArturKlajnerok: Solved in: 44beb1f

PR #27, opened 2015-05-11
[SERVICES-315] Fix for parsing null from user_editcount

ArturKlajnerok:
As user out users can have a `NULL` value in `user_editcount` field, we need to parse it to a pointer, otherwise `sql: Scan error on column index 13: converting string "<nil>" to a int64: strconv.ParseInt: parsing "<nil>": invalid syntax` error will be returned.

/cc @Wikia/services-team


nmonterroso:
LGTM

PR #28, opened 2015-05-12
return all validation errors

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-275

Modifies the user registration flow to pass around a `map[string][]error` containing all the validation errors the proposed username, email, password, and birthdate contain. Clients will see:

```
error_description: {
  username: [
    error1,
    error2
  ],
  password: [
    error1,
    error2
  ]
}
```


helios/register_controller.go
```diff
@@ -187,18 +217,44 @@ func (controller *RegisterController) setOutputSuccess(resp *osin.Response, desc
 	resp.Output["success_description"] = description
 }
 
-func (controller *RegisterController) setOutputError(resp *osin.Response, message, description string) {
+func (controller *RegisterController) setOutputError(resp *osin.Response, message string, requestErrors map[string][]string) {
 	resp.Output["error"] = message
-	resp.Output["error_description"] = description
+	resp.Output["error_description"] = requestErrors
```
nmonterroso: spoke with @ArturKlajnerok and since there are no Helios clients in the wild that are the authority for registering users, it's ok to change `error_description`'s type from `string` to `map[string][]string`

utils/utils.go
```diff
@@ -15,6 +15,10 @@ import (
 
 const (
 	InfluxSeriesName = "metrics"
+	UsernameField    = "username"
+	PasswordField    = "password"
+	EmailField       = "email"
+	BirthdateField   = "birthdate"
```
ArturKlajnerok: I would move the "Fields" const to models/user.go as they seem to fit there more :panda_face:

helios/register_controller.go
```diff
@@ -90,22 +98,43 @@ func (controller *RegisterController) registerHandler(w http.ResponseWriter, r *
...
+func (controller *RegisterController) checkPhalanx(
+	name string,
+	validationErrors map[string][]error) (bool, map[string][]error) {
+
+	isPhalanxBlocked, blockedError := controller.isPhalanxBlocked(name)
+	isAntiSpoof, spoofError := controller.isAntiSpoof(name)
+	hasError := false
+
+	if blockedError != nil {
+		validationErrors[utils.UsernameField] = append(validationErrors[utils.UsernameField], blockedError)
```
drsnyder: I see `validationerrors` reference twice here. Does append modify the reference to `validationErrors`?

nmonterroso: `append` returns a new `slice`, it does not modify it in place: https://blog.golang.org/slices

drsnyder: Should this be an abstraction? Something like:

```go
errors.addUsernameError(blockedError)
```

Currently the controller needs to know about how the errors are stored (a map) and the field in the map.

nmonterroso: Thought about that as I was writing it, but I don't think that's worth the effort since it will be an abstraction specific to this use case, since as you noted it will have to know that it's adding to the `username` list of errors.

utils/utils.go
nmonterroso: moved in 2097279

helios/register_controller.go
```diff
@@ -90,22 +98,43 @@ func (controller *RegisterController) registerHandler(w http.ResponseWriter, r *
...
+}
 
-	if blocked := controller.checkPhalanxBlock(resp, user.Name); blocked {
-		return
+func (controller *RegisterController) checkPhalanx(
+	name string,
+	validationErrors map[string][]error) (bool, map[string][]error) {
+
+	isPhalanxBlocked, blockedError := controller.isPhalanxBlocked(name)
+	isAntiSpoof, spoofError := controller.isAntiSpoof(name)
```
drsnyder: These seem out of place to me in that it doesn't seem like it's the job of the controller to determine if the name is blocked (specifically by phalanx) or anti-spoof. I would expect to find these in a service or utilities package.

What does it mean for a username to be anti-spoof? Does the name of that method reflect what it's doing?

helios/register_controller.go
```diff
@@ -90,22 +98,43 @@ func (controller *RegisterController) registerHandler(w http.ResponseWriter, r *
...
+	if len(validationErrors) > 0 {
+		controller.setValidationErrors(resp, validationErrors, hasPhalanxServiceError)
+	} else {
+		controller.createAccount(resp, user)
 	}
+}
 
-	if blocked := controller.checkPhalanxBlock(resp, user.Name); blocked {
-		return
+func (controller *RegisterController) checkPhalanx(
```
drsnyder: Can we change the name of this to reflect it's function and not the service it's connecting to? Is this `isValidUsername`?

drsnyder:
Unit tests?

nmonterroso:
There are no unit tests for Helios, just end-to-end tests, which were added.

helios/register_controller.go
nmonterroso: Kinda but not really. It's checking if it's valid in the sense that it isn't a blocked username, or if it's "spoofed".

drsnyder:
There [are](https://github.com/Wikia/helios/blob/master/utils/utils_test.go) [some](https://github.com/Wikia/helios/blob/master/helios/oauth_controller_test.go), but not enough. IMO we need to do a better job of increasing the unit test coverage in helios.

helios/register_controller.go
drsnyder: `isSpoofedUsername`? What does "spoofed" mean?

helios/register_controller.go
nmonterroso: I don't get what you mean by having these in a service or utilities package. The service is Phalanx, which is queried using https://github.com/Wikia/go-commons/blob/master/phalanx/phalanx.go.

nmonterroso:
Hm interesting, I didn't see those. I'll add some.

helios/register_controller.go
nmonterroso: Also, re: spoof. I'm not really sure, but I spoke with @ArturKlajnerok and some of it has to do with trying to create an account whose name could be confused with another. Example if my username is `nelson` with an "L", it should prevent creation of an account with the username `neIson` with a capital "i". I think it's one of the requirements for usernames at registration-time, so Helios needs to hook into it somehow.

helios/register_controller.go
nmonterroso: I'll come up with a better name, `isSpoofed` isn't correct either, since it's a wrapper for calling both phalanx functions.

nmonterroso: Also, I added some info on what `spoof` means in an above comment.

helios/register_controller.go
drsnyder: To me, controllers should be thin. They are just passing messages from the outside world to internal libraries, services, and models, gathering the response and sending it back.

I think it's a step in the right direction to move this into `checkPhalanx` but I think we could go one step forward to have something like `spamService.isSpammer`. If you assume that controllers should be thin, then another way to think about this is why does a controller need a client to the anti spam service? Shouldn't that be someone else's responsibility?

Or, why should we have to change controller code if we changed the anti-spam service from `phalanx` to `prophylactic`? Or what if we changed `phalanxClient.Check()` to return a map of error messages? Would it make sense that we would also have to change controller code?

nmonterroso: That's what it's doing, though. The controller is passing the request to the phalanx library and reacting to the result by changing the HTTP status code/error response if necessary.

In your example, at first glance we wouldn't have to modify the controller code, but we'd have to modify the `spamService` code. Then, what would we do with the map of error messages there?   Assuming we still want to pass all the error messages, the controller would still need them, so it would have to process them anyway.

models/user.go
```diff
@@ -197,12 +178,13 @@ func (user *User) IsValidPassword(password string) bool {
 	return user.HashedPassword == OldHashPassword(password, user.Id)
 }
 
-func parseBirthDate(birthDate string) (*time.Time, error) {
+func parseBirthDate(birthDate string, validationErrors map[string][]error) (*time.Time, map[string][]error) {
 	date, err := time.Parse(BirthDateFormat, birthDate)
 	if err != nil {
-		return nil, err
+		validationErrors[BirthdateField] = append(validationErrors[BirthdateField], ErrBirthDateNotValid)
```
ArturKlajnerok: I think you don't need to set this error here as if parsing will fail and nil will be returned, the error will be set here: https://github.com/Wikia/helios/pull/28/files#diff-46259196476f860fea33754fcb22e9eeR147

nmonterroso: Ah, good catch. I didn't like doing this but at the time I thought there was no way to avoid it. Updated in b5bf8c3, and now the `Validate()` function creates the error map, which feels more correct.

helios/register_controller.go
drsnyder: > That's what it's doing, though. The controller is passing the request to the phalanx library and reacting to the result by changing the HTTP status code/error response if necessary.

Yes. What I'm suggesting is that some of that logic doesn't belong in the controller. The controller shouldn't be concerned with calling the phalanx service and interpreting the result. It should call an anti-spammer service or helper that has a clear contract and then alter the response accordingly. 

> In your example, at first glance we wouldn't have to modify the controller code, but we'd have to modify the spamService code. Then, what would we do with the map of error messages there? Assuming we still want to pass all the error messages, the controller would still need them, so it would have to process them anyway.

I'm imagining something like this:

```go
// returns bool, map of errors
isBlocked, validationErrors := controller.spamService.isSomethingSpamRelatedWeDontWant(name)
if isBlocked {
    // status is 400, create Problem response and add validationErrors
}
```

The `controller.spamService` is responsible for all the nuts and bolts of checking whether the name is blocked or not. This includes details that the controller shouldn't care about such as what the spam service is and what it returns.

Even though the change is small I think the advantages are that
 * It's easier to test. If you want to test `registerHandler` you do so more effectively because you can inject the `spamService` and mock at a higher level of abstraction.
 * There is less coupling between the controller and phalanx and higher [cohesion](http://en.wikipedia.org/wiki/Cohesion_%28computer_science%29) in the controller.

nmonterroso: I aim to (somewhat) please: 0d764f0

nmonterroso:
@drsnyder added unit tests in 29b0217e484cc8c59a99b80b1c2ac644ba7a5f93

nmonterroso:
good call @ArturKlajnerok 7db811c

service/user_registration_external_validator.go
```diff
@@ -0,0 +1,80 @@
...
+	serverConfig  *config.ServerConfig
+	phalanxClient *phalanx.Client
+	wikiaClient   *wikia.Client
+}
+
+var (
+	ErrExternalValidation = errors.New("Username is unavailable")
+)
+
+func NewUserValidator(serverConfig *config.ServerConfig) *UserRegistrationExternalValidator {
```
drsnyder: Should this be `NewUserRegistrationExternalValidator`?

drsnyder:
:+1:

service/user_registration_external_validator.go
nmonterroso: Hm, maybe, but isn't `New` implied when registering? Don't think we'll ever have `ExistingUserRegistrationExternalVaidator` :P

PR #30, opened 2015-05-13
remaining registration checks

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-280

Adding phalanx check for email field and returning descriptive message when a username is taken. We were previously returning a `500: <something is broken>`


service/user_registration_external_validator.go
```diff
@@ -46,6 +47,20 @@ func NewUserValidator(serverConfig *config.ServerConfig) *UserRegistrationExtern
 }
 
 func (validator *UserRegistrationExternalValidator) Validate(user *models.User) (bool, error) {
+	validUsername, err := validator.validateName(user)
+	if err != nil {
```
aquilax: There is no need of calling `validateEmail` if `validUsername` is false. This can become:
```go
if err != nil || !validUsername {
	return false, err
}
``

drsnyder:
lgtm

PR #33, opened 2015-05-15
Storage refactor

nmonterroso:
@Wikia/services-team 

moving database connection stuff from the `StorageFactory` to `UserStorage` and updating `StorageFactory` to be a map for the `main` storage (when not in shadow mode) and the `shadow` storage (when in shadow mode). Also adding a new config `user-table-shadow` which the `shadow` `UserStorage` will use. This allows us to control authentication and registration shadow mode independently.


ArturKlajnerok:
Lgtm :+1:

PR #34, opened 2015-05-15
Update the hash for the apiclient in godeps

drsnyder:
This moves the hash to the current HEAD in go-commons for the apiclient package. This will pull in the `request.Close = true` change.


ArturKlajnerok:
Lgtm :+1:

PR #35, opened 2015-05-18
don't allow new user creation if email is taken

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-157

As part of this, I abstracted the management of the `map` of `error`s when validating a user to `common.ErrorList` and used that in `User.CreateUser` to account for the possibility of a username and an email already belonging to a user.


controllers/register_controller.go
```diff
@@ -177,6 +177,7 @@ func (controller *RegisterController) setRequestErrors(resp *osin.Response, requ
 }
 
 func (controller *RegisterController) setServerError(resp *osin.Response, err error) {
+	println("error", err.Error())
```
drsnyder: Did you mean to leave this in?

nmonterroso: OOPS

nmonterroso: removed in e1918f4

common/error_list.go
```diff
@@ -0,0 +1,32 @@
+package common
+
+type ErrorList struct {
+	errors map[string][]error
+}
```
drsnyder: I'm just thinking out loud and trying to learn more go in the process. Don't let this gate merging.

Should this be an interface? Something like:

```go
type ErrorList interface {
   IsEmpty() bool
   Add()
   AsStrings() map[string][] string
}

type HeliosErrorList map[string][]error

func (list *HeliosErrorList) IsEmpty() bool {
   return len(list) == 0
}

func (list *HeliosErrorList) Add(key string, err error) {
   list[key] = append(list[key], err)
}

// probably not necessary since I think you can create it directly
func NewErrorList() *HeliosErrorList {
   return HeliosErrorList(make(map[string][]error)
}
...
```

That would at least allow you to stub/mock the functionality in test that didn't bind you to a specific implementation. I'm thinking about this as a more general practice after reading [this post about testing in go](http://relistan.com/writing-testable-apps-in-go/) and the [effective go section on interfaces and type](https://golang.org/doc/effective_go.html#interfaces_and_types).

nmonterroso: Hm, maybe. Though I'd say that in your example `HeliosErrorTest` should be a struct. I tried this, and

```go
list[key] = append(list[key], err)
```

causes the error `invalid operation: test["key"] (type *ErrorListTest does not support indexing)`. There's probably a way to do this correctly, but I'd still prefer to have the data structure (`map[string][]error`) behind a struct to avoid something from the outside from modifying it directly. I do see the value in using an `interface`, though.

drsnyder: Ok, cool.

nmonterroso: Though, thinking about it a bit more I think it's OK to leave as-is for now, since the implementation is a simple data structure, and any stub implementation in a test would be nearly identical to the actual implementation.

nmonterroso: Also, added some tests in 8cfe8f8

drsnyder:
:+1: you read my mind

models/user_storage.go
```diff
@@ -66,8 +80,18 @@ func (userStorage *UserStorage) Close() {
 	userStorage.slaveMap.Db.Close()
 }
 
-func (userStorage *UserStorage) FindByName(userName string, mustExist bool) (*User, error) {
+func (userStorage *UserStorage) EmailInUse(email string, mustExist bool) (bool, error) {
+	query := fmt.Sprintf("select count(*) from %s where user_email=?", userStorage.tableName)
```
drsnyder: @nmonterroso sorry, I didn't catch this until I took a closer look at the user model. I don't think we should be doing this at the application layer if we want to ensure that this constraint is maintained.

Practically we can't put a `UNIQUE` constraint on `user_email` because there are already duplicates. In fact it looks like in dev we set all the emails to `devbox@wikia-inc.com`.

As an alternative, we might be able create a trigger on insert that checked to make sure the email doesn't exist. If we can't do that then we probably want to wrap this in a transaction.

nmonterroso: Ooh, yeah that sounds like a good idea

PR #36, opened 2015-05-18
[SERVICES-207] Username validation resource

ArturKlajnerok:
Resource for validation and checking the availability of username. Supports:
* Restricted characters validation
* Min/Max length validation
* AntiSpoof check
* Phalanx check
* Username availability in user DB

Additionally moved helios to server package and did some small tests cleanup. 

/cc @Wikia/services-team


controllers/register_controller.go
```diff
@@ -103,6 +104,52 @@ func (controller *RegisterController) registerHandler(w http.ResponseWriter, r *
...
 
+// @Title Register
+// @Description Username validation
+// @Accept  json
+// @Param   username      form string true "Username for validation"
+// @Failure 400 {object} string "Invalid input"
+// @Router /username/validation [post]
+func (controller *RegisterController) usernameValidationHandler(w http.ResponseWriter, r *http.Request) {
+	timer := utils.CreateTimerForAPICall(controller.influxdbClient, common.AppName, "usernameValidationHandler")
+	defer utils.CloseTimer(timer)
```
drsnyder: This looks like middleware/request wrapper. Should there be a utility function that's something like:

```go
func timedHandler(controller *TimedController, appName string, handlerName string, fn func(http.ResponseWriter, *http.Request)) http.HandlerFunc {                                                                                                                                                                                                                                                                                                                                                    
  return func(w http.ResponseWriter, r *http.Request) {                                                                                                                                                                                                                                                                                                                                                                               
    timer := utils.CreateTimerForAPICall(controller.MetricClient(), appName, handlerName)
    defer utils.CloseTimer(timer)                                                                                                                                                                                                                                                                                                                                                                                                        
    fn(w, r)                                                                                                                                                                                                                                                                                                                                                                                                                          
  }                                                                                                                                                                                                                                                                                                                                                                                                                                   
}
...
http.HandleFunc("/username/validation", timedHandler(controller, common.AppName, "usernameValidationHandler", originHandler(controller.usernameValidationHandler))
```

This would require the addition of an interface that supported `MetricClient` to be generalized.

controllers/register_controller.go
```diff
@@ -103,6 +104,52 @@ func (controller *RegisterController) registerHandler(w http.ResponseWriter, r *
...
+
+	resp := controller.server.NewResponse()
+	defer resp.Close()
+	defer osin.OutputJSON(resp, w, r)
```
drsnyder: It seems odd that we have a dependency on `osin` here when this isn't related to OAuth. Do we have to use `osin` for JSON generation?

controllers/register_controller.go
ArturKlajnerok: Yeah, I was thinking about generalizing this also with other handler code, as each of our handlers should have : 
 ```if origin := r.Header.Get("Origin"); origin != "" {
			w.Header().Set("Access-Control-Allow-Origin", origin)
		}```

controllers/register_controller.go
```diff
@@ -103,6 +104,52 @@ func (controller *RegisterController) registerHandler(w http.ResponseWriter, r *
...
+
+	err := r.ParseForm()
+	if err != nil {
+		controller.setRequestError(resp, "request", controller.deferror.Get(osin.E_INVALID_REQUEST))
+		return
+	}
+
+	user := &models.User{Name: r.Form.Get(models.UsernameField)}
+	user.NewValidationErrorMap()
```
drsnyder: This seems out of place. It also looks like this is done in `NewUser` so why should the caller have create their own `User` struct and initialize some of the internals?

models/user.go
```diff
@@ -79,19 +79,23 @@ func NewUser(name, password, email, birth string) *User {
...
-	user.validationErrors = make(map[string][]error)
 
+	user.NewValidationErrorMap()
 	user.validate()
 
 	return user
 }
 
+func (user *User) NewValidationErrorMap() {
+	user.validationErrors = make(map[string][]error)
```
nmonterroso: This has recently been changed in [master](https://github.com/Wikia/helios/blob/master/models/user.go#L82) to use the new [`ErrorList`](https://github.com/Wikia/helios/blob/master/common/error_list.go) struct. Can you merge in master and use that? Also this function might be better named as `ResetErrorList`

controllers/register_controller.go
```diff
@@ -103,6 +104,52 @@ func (controller *RegisterController) registerHandler(w http.ResponseWriter, r *
...
+func (controller *RegisterController) usernameAvailable(resp *osin.Response, user *models.User) {
+	user.MediaWikiTransformation()
+	existing, err := controller.userStorage.FindByName(user.Name, false)
+	if err != nil {
+		controller.setServerError(resp, err)
+	} else if existing != nil {
+		controller.setRequestError(resp, models.UsernameField, models.ErrUserExists.Error())
+	} else {
+		controller.setOutputSuccess(resp, "Username available")
+	}
```
drsnyder: This pattern is repeated 3-4 times in this controller. Should we try and consolidate it into a helper? By pattern I mean:

```go
if err != nil {
    // server error
} else if <other test> {
   // request error
} else {
   // success
}
```

controllers/register_controller.go
drsnyder: Yeah, I think it would be better to separate these so that each function/wrapper does one thing. If we want a higher level wrapper that combines the three I think that's fine.

Also, is there a reason we are using `osin.OutputJSON(resp, w, r)` for all of our JSON output?

ArturKlajnerok: We where using it with OAuth controller, but now I don't see any more specific reason for using it with other controllers.

controllers/register_controller.go
ArturKlajnerok: It's used for convince of using [Response](https://github.com/RangelReale/osin/blob/6ed440e2fa355d0ea2163553b2e1c8a495a0ce01/response.go) ] and not implementing similar struct as it's available with lib.

models/user.go
```diff
@@ -79,19 +79,23 @@ func NewUser(name, password, email, birth string) *User {
...
 	user.EditCount = getDefaultEditCount()
-	user.validationErrors = common.NewErrorList()
 
+	user.NewValidationErrorList()
 	user.validate()
 
 	return user
 }
 
+func (user *User) NewValidationErrorList() {
```
nmonterroso: do we still need this function?

models/user.go
ArturKlajnerok: Changed to ErrorList here: 0cb5475, as for name it's creating fresh new ErrorList in User so for me the current one make sense.

models/user.go
ArturKlajnerok: Used here : https://github.com/Wikia/helios/pull/36/files#diff-02d58c545e9b145ed19d6f39f7c1c37fR124

nmonterroso:
LGTM - anything else from your side @drsnyder ?

ArturKlajnerok:
@drsnyder Changes with the middleware will be part of separate PR

drsnyder:
LGTM, we can fix the middleware and `/users` resource in follow on PRs

ArturKlajnerok:
@drsnyder Timed handler middlewere added here : #39

PR #37, opened 2015-05-18
[SERVICES-31] Removed access restrictions with client credentials

ArturKlajnerok:
Considering rate limiting of register resource, restricting access with client credentials is not needed anymore.

/cc @Wikia/services-team


nmonterroso:
Will we need to re-add this if/when we implement oauth?

ArturKlajnerok:
@nmonterroso If we will go with full oauth2, we will restrict access to this resource based on tokens, so adding credentials here was a temporary solution.

nmonterroso:
ah ok. LGTM.

PR #38, opened 2015-05-21
Everyone's Got Problems (and error codes)

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-318

Changes the English error-messages to be descriptive error code strings. It'll be up to clients to turn those codes into user-facing translated strings. Also introducing the `Problem` struct as defined in the story, and the `Problematic` interface to convert other types to `Problem`, which `ErrorList` now implements. 

Note that we cannot change the `Content-Type` header right now because the osin library forcefully adds `Content-Type: application/json` to responses, and we'd end up with multiple `Content-Type` headers. See https://github.com/RangelReale/osin/pull/60


common/error_list.go
```diff
@@ -1,9 +1,17 @@
...
+	"github.com/Wikia/helios/problem"
+)
 
 type ErrorList struct {
 	errors map[string][]error
 }
 
+const (
+	AdditionalField = "field"
+)
```
drsnyder: Does `error_list` need to know about this? I would expect something like `aProblem.AddAdditional(field)` instead of `aProblem.AddAdditional(AdditionalField, field)`.

common/error_list.go
```diff
@@ -54,3 +62,23 @@ func (list *ErrorList) AsStrings() map[string][]string {
 
 	return errorMessages
 }
+
+func (list *ErrorList) AsProblem(title string) *problem.Problem {
+	errors := make([]*problem.Error, 0)
+
+	for field, errorList := range list.errors {
+		for _, error := range errorList {
+			errors = append(errors, problem.NewError(error.Error()).Additional(AdditionalField, field))
```
drsnyder: Is the `error.Error()` required? That is, can this be `problem.NewError().Additional(field)`?

controllers/helpers.go
```diff
@@ -13,3 +15,10 @@ func originHandler(fn func(http.ResponseWriter, *http.Request)) http.HandlerFunc
 		fn(w, r)
 	}
 }
+
+func setOutputProblem(resp *osin.Response, p problem.Problematic, problemTitle string, statusCode int) {
+	// commented out since the osin library always adds content-type application/json https://github.com/RangelReale/osin/pull/60
+//	resp.Headers.Set("Content-Type", problem.ContentType)
+	resp.Output = p.AsProblem(problemTitle).AsMap()
+	resp.StatusCode = statusCode
+}
```
drsnyder: I think we should decouple our output from `osin`. It's convenient that it already provides an output generator but it's also coupling us to a 3rd party library that a) we might swap out later, b) we might remove entirely c) whose interface may change.

common/error_list.go
nmonterroso: It does need to know about it. An additional is a map of key/vals that are specific to the type of error that's being reported. For these the map key is "field" and the value is the field of registration that failed, like "username" or "email". A different problem type could be something like:

```json
{
  "code": "something_went_wrong",
  "additional": {
    "count": 5,
    "other_key": "something else"
  }
}
```

Though maybe doesn't belong in `ErrorList` but in `User`

common/error_list.go
nmonterroso: It could be, though it's never accessed as an `error` once it gets into the `problem.Error` struct, so I don't see value in doing that right now.

controllers/helpers.go
nmonterroso: That sounds right, but that should probably be part of its own PR. Can you open a ticket to do that?

drsnyder: Yeah, I don't think it should be done as part of this PR. It feels a little like we are cobbling together a custom web framework in helios.

nmonterroso:
@drsnyder I added `errorlist.UserValidation` which subtypes `ErrorList` and overwrites the `AsProblem` function, so now `ErrorList` doesn't know any details about the `additional` required when validation a user. Can you take a look?

controllers/register_controller_e2e_test.go
```diff
@@ -165,3 +155,18 @@ func TestE2eRegisterTestSuite(t *testing.T) {
...
+		if additional[errorlist.ValidationAdditionalField] == field {
+			return true
+		}
+	}
+
+	return false
+}
+
+func pseudoRandomString(base string, purpose string) string {
+	return base + purpose + strconv.FormatInt(time.Now().Unix(), 10)
```
drsnyder: Minor: Want to use `rand.Int63()` from `math/rand`?

problem/problem.go
```diff
@@ -0,0 +1,63 @@
...
+
+	for _, error := range problem.errors {
+		errors = append(errors, error.AsMap())
+	}
+
+	return map[string]interface{} {
+		Description: problem.description,
+		Errors: errors,
+	}
+}
```
drsnyder: We probably mean `title` in here rather than `description`. The `description` field is not in the [draft](https://tools.ietf.org/html/draft-nottingham-http-problem-07#section-3.1).

errorlist/error_list.go
```diff
@@ -54,3 +59,23 @@ func (list *ErrorList) AsStrings() map[string][]string {
...
+	errors := make([]*problem.Error, 0)
+
+	for _, errorList := range list.errors {
+		for _, error := range errorList {
+			errors = append(errors, problem.NewError(error.Error()))
+		}
+
+	}
+
+	return problem.NewProblem("An unexpected error occurred", errors)
```
drsnyder: :+1: we are one step closer to symmetry with the [Java interface in pandora](https://github.com/Wikia/pandora/blob/b12fa2a62869e795b9a35f385d553b62531eece0/lib/service-core/src/main/java/com/wikia/pandora/http/StandardResponses.java.

drsnyder: Not that symmetry should necessary be a goal, but I think it will lower the cognitive burden when switching back and forth.

controllers/register_controller_e2e_test.go
nmonterroso: 03f0a4b

controllers/helpers.go
```diff
@@ -13,3 +15,10 @@ func originHandler(fn func(http.ResponseWriter, *http.Request)) http.HandlerFunc
 		fn(w, r)
 	}
 }
+
+func setOutputProblem(resp *osin.Response, p problem.Problematic, problemTitle string, statusCode int) {
```
drsnyder: :boom: now can have helpers like `NotFound()`, `Conflict()`, etc.

problem/problem.go
nmonterroso: 8746829

problem/problem.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (problem *Problem) AsMap() map[string]interface{} {
+	errors := make([]map[string]interface{}, 0)
+
+	for _, error := range problem.errors {
+		errors = append(errors, error.AsMap())
+	}
+
+	return map[string]interface{} {
+		Description: problem.title,
```
drsnyder: Should this be `Title`?

nmonterroso: oops af205f9

controllers/register_controller_e2e_test.go
nmonterroso: Actually, these ints make the string too big, so gonna stick with time for now.

drsnyder:
:sheep: it

ArturKlajnerok:
Reviewed it after merge and LGTM :smile_cat:

PR #40, opened 2015-05-22
Introduce a "user" domain object

drsnyder:
Introduce a "user" domain object and decouple it from the database layer. Currently, the `models.User` object conflates validation and business plus database object.

This PR introduces a `domain.User` interface and a `domain.UserValue` (practically immutable) struct and moves the essential "user" business logic (uppercase first letter, birthday formatting, etc) to the domain object.

The next step will be move the validation from `models.User` to `domain.User`. I would like some feedback on this part. What I'm considering is something like the following from the top down:

```go
// /user resource
user := createFromForm(request)

userValidator = validation.NewUserValidator(validation.AllValidationFuns...)
valid, errors := userValidator.Validate(user) 

if valid {
  // create user, return ok
} else {
   return Response.BadRequest("The provided user is invalid. See the errors for details.", errors)
}
```

/cc @Wikia/services-team 

TODO:
 * [x] Fix the tests


drsnyder:
One subtlety that this refactoring surfaced was that [this test (`user.Password == user.Name`)](https://github.com/Wikia/helios/blob/master/models/user.go#L140) probably never be true under normal usage. This is because we always make the first letter of the username uppercase.

In the `register` controller the order of operations of mutating the name is different depending on the context. Compare [here](https://github.com/Wikia/helios/blob/master/controllers/register_controller.go#L141-143) vs [here](https://github.com/Wikia/helios/blob/master/controllers/register_controller.go#L160-161). In the former we validate the user, then we mutate the name but it's the reverse in the latter.

IMO, whenever we reference the username it should *always* be uppercase. This is business logic that is central to the "user" domain. To this end I moved it into the [user domain object](https://github.com/Wikia/helios/pull/40/files#diff-ce64e45f7310785b2dff83b92bf36c4aR47). I also suggest that the test for the username being equal to the password should be done on strings that have been normalized to lower case.

@Wikia/services-team any objections?

ArturKlajnerok:
I agree that validation should be consistent and making username always start from uppercase will help to achieve that, let's just make it visible that it's a MW specific transformation. As for normalizing password and username to lower case for comparison, I think it's a good improvement and we just need to be aware that this will produce different result than MW validation. :+1:

errorlist/error_list.go
```diff
@@ -1,32 +1,49 @@
...
+	Key     string
+	Message string
+}
+
+func (error *ValidationError) Error() string {
+	return error.Message
+}
+
+func NewValidationError(key string, err string) *ValidationError {
+	validationError := &ValidationError{key, err}
```
drsnyder: @nmonterroso I think this form is ok. In this case, because the struct contains primitives, it can be initialized directly without `new`.

errorlist/error_list.go
```diff
@@ -48,34 +65,44 @@ func (list *ErrorList) ContainsInKey(key string, err error) bool {
...
 
-func (list *ErrorList) AsProblem() *problem.Problem {
+func (list *ErrorList) AsProblem(title string) *problem.Problem {
 	errors := make([]*problem.Error, 0)
 
-	for _, errorList := range list.errors {
+	for field, errorList := range list.errors {
 		for _, error := range errorList {
-			errors = append(errors, problem.NewError(error.Error()))
+			errors = append(errors, problem.NewError(error).Additional(ValidationAdditionalField, field))
```
drsnyder: This was the primary use case for the `UserValidation` object. I moved it here and removed that struct.

models/user.go
```diff
@@ -66,113 +41,47 @@ type User struct {
 	EditCount           *int64                    `db:"user_editcount"`
 	BirthDate           *time.Time                `db:"user_birthdate"`
 	Options             string                    `db:"user_options"`
-	validationErrors    *errorlist.UserValidation `db:"-"`
+	// FIXME: remove once the validation is decoupled
+	validationErrors    *errorlist.ErrorList      `db:"-"`
 }
 
-func NewUser(name, password, email, birth string) *User {
+func FromDomain(domainObject domain.User) *User {
```
drsnyder: I think going forward, we should only create `models.User` objects from the domain object. That will provide the interface between the domain and the storage layers.

drsnyder:
Ok, this is ready for a second review @nmonterroso @ArturKlajnerok.

controllers/register_controller.go
```diff
@@ -94,20 +98,20 @@ func (controller *RegisterController) registerHandler(w http.ResponseWriter, r *
...
-		r.Form.Get(models.UsernameField),
-		r.Form.Get(models.PasswordField),
-		r.Form.Get(models.EmailField),
-		r.Form.Get(models.BirthdateField))
-	user.Validate()
-	_, err = controller.externalValidator.Validate(user)
+	user := createDomainUserFromForm(r)
+	isDomainObjectValid, userValidationErrors := controller.userValidator.Validate(user)
+
+	// FIXME: refactor the external validator to be a user validator-- see the interface in validation
```
nmonterroso: Can we do this before we merge? Or is this the technical debt you were saying change would introduce? If so, let's get a ticket in the backlog so we can avoid having `TODO`s and `FIXME`s in Helios

drsnyder: > Can we do this before we merge?

We could, but I don't want to slow down our deliverables. This took me a half a day or more longer than I expected. 

I'm still not 100% sure how to handle a validator that has external dependencies. For example, would the following work:

```go
func (service *UserRegistrationExternalValidator) ValidateName(user domain.User) (bool, *errorlist.ValidationError) {  
...
}
controller.userValidator.Add(externalValidator.ValidateName)
sDomainObjectValid, userValidationErrors := controller.userValidator.Validate(user)
```

I think so, but I haven't tried it. If you are ok with it, I'll create the tickets for the FIXME's.

domain/user.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/Wikia/helios/common"
+	"github.com/Wikia/helios/utils"
+)
+
+const (
+	BirthDateFormat = "2006-01-02"
+	SaltSize        = 8
+)
+
+type Hashable interface {
```
nmonterroso: Does anything implement this interface? Should there also be something like

```go
type Password string

func (string String) Hash() string {
  ...
}

// and other Hashable functions on Password
```

controllers/register_controller.go
drsnyder: https://wikia-inc.atlassian.net/browse/SERVICES-349

errorlist/error_list.go
nmonterroso: dope

models/user.go
```diff
@@ -66,113 +41,50 @@ type User struct {
...
-}
-
-func (user *User) MediaWikiTransformation() {
-	user.Name = utils.FirstToUpper(user.Name)
+func (user *User) IsValid() bool {
+  return user.validationErrors.IsEmpty()
 }
 
+// FIXME: This is not really validation but authentication.
 func (user *User) IsValidPassword(password string) bool {
```
drsnyder: I created https://wikia-inc.atlassian.net/browse/SERVICES-350 for this. The FIXME's below are related.

errorlist/error_list.go
```diff
@@ -48,34 +65,44 @@ func (list *ErrorList) ContainsInKey(key string, err error) bool {
...
+	return list.errors
+}
+
+func (list *ErrorList) Merge(mergeWith *ErrorList) {
+	for k, v := range mergeWith.Errors() {
+		list.errors[k] = append(list.errors[k], v...)
+	}
+}
+
 func (list *ErrorList) AsStrings() map[string][]string {
```
nmonterroso: This function can probably just return `list.errors` now, I don't think it's doing anything other than creating a new struct and copying the contents of `list.errors` into it. Or, if we want to maintain that we don't expose `list.errors` directly for modification?

errorlist/error_list.go
nmonterroso: This is the way it was before :(

We added the `UserValidation` struct so `ErrorList` could remain a generic container for a list of errors and wouldn't have to know about user validation `Additional`s at all - since `Additional` is a map of key/vals that are specific to the type of `Error` that's being added.

models/user_storage.go
```diff
@@ -47,21 +46,23 @@ func NewUserStorage(table string, dbConfig *config.DbConfig) *UserStorage {
...
+		validationErrors.Add(ErrUsernameExists)
 	}
 
 	emailInUse, err := userStorage.EmailInUse(user.Email, false)
 	if err != nil {
 		return validationErrors, err
 	} else if emailInUse {
-		validationErrors.Add(UsernameField, ErrEmailExists)
+		// FIXME: move this to the validation logic
+		validationErrors.Add(ErrEmailExists)
```
drsnyder: These two could also be addressed by https://wikia-inc.atlassian.net/browse/SERVICES-349.

models/user.go
```diff
@@ -66,113 +41,50 @@ type User struct {
...
 	user.EditCount = getDefaultEditCount()
-	user.validationErrors = errorlist.NewUserValidation()
-
-	return user
-}
+	user.validationErrors = errorlist.NewErrorList()
 
-func (user *User) AddValidationError(field string, err error) {
-	user.validationErrors.Add(field, err)
+  return user
```
nmonterroso: spacing off here?

errorlist/error_list.go
drsnyder: It's not being used. Removed in e82af4c.

models/user_storage.go
nmonterroso: cool

controllers/register_controller.go
nmonterroso: Cool, I'm OK with having `FIXME`s as long as we have the stories created in JIRA :+1:

models/user.go
drsnyder: Fixed in 61193a4.

common/constants.go
```diff
@@ -2,8 +2,14 @@ package common
 
 
 const (
+	MinAge      = 13
+	MinLen      = 1
+	MaxLen      = 50
+	MinHashSize = 32
```
ArturKlajnerok: Why in common if used only in model package ?

domain/user.go
drsnyder: No, I don't think it belongs here. It should be part of the refactoring for https://wikia-inc.atlassian.net/browse/SERVICES-350. I removed it in b2f1ed2.

errorlist/error_list.go
drsnyder: Ok, but currently this is the only kind of `ErrorList` we are creating. If we need something more generic we can separate that out. My guess is that we'll probably only have `Problem` and `ErrorList`s that convert to `Problem`s with the additional `errors` member.

common/constants.go
drsnyder: That was probably overly aggressive or preemptive reorganizing on my part. Should I put it back?

ArturKlajnerok: I would keep const that are used only in on package in the package :panda_face:

drsnyder: Ok, fixed in a8a909c.

ArturKlajnerok:
I agree we should address all fixme proposals, other then that LGTM :+1:

nmonterroso:
I think we'll eventually have to move the `Validation` references outside of `ErrorList`, but if we're OK with leaving that as a task for Future Damon then LGTM

drsnyder:
> I think we'll eventually have to move the Validation references outside of ErrorList, but if we're OK with leaving that as a task for Future Damon then LGTM

Yes, that part is still a little awkward. But for now I think it's passable since the only kind of `ErrorList` we have is one used for validation.

PR #42, opened 2015-05-27
[SERVICES-182] Facebook login connect new account

ArturKlajnerok:
Facebook login & Wikia user creation flow works in this manner: 
* Client that's already logged in with Facebook makes an request to login user in Wikia with provided fb access token
* If there's no Wikia and Facebook users mapping proper response with that information is returned
* Additional data are required to be passed for user creation as username and password
* Client makes an request to create Wikia user based on additional data and basic Facebook data, also mapping with Wikia and Facebook users need to be done.

- [x] Facebook graph api support 
- [x] Get Facebook user from graph api
- [x] Check if Facebook account is not already mapped with Wikia account
- [x] Create a valid access token for Facebook account is mapped with Wikia account
- [x] Create Wikia user account
- [x] Create Facebook and Wikia user mapping

Please provide feedback @Wikia/services-team


controllers/facebook_controller.go
```diff
@@ -0,0 +1,65 @@
...
+	userStorage    *models.UserStorage
+}
+
+func NewFacebookController(influxdbClient *client.Client, storageFactory *models.StorageFactory) *FacebookController {
+
+	controller := new(FacebookController)
+	controller.influxdbClient = influxdbClient
+	controller.userStorage = storageFactory.GetUserStorage(models.MainStorage)
+
+	http.HandleFunc("/facebook/login",
```
drsnyder: Is this equivalent to `/token` under the normal flow? If so should we use the same terminology?

models/user_facebook.go
```diff
@@ -0,0 +1,13 @@
+package models
+
+import (
+	"time"
+)
+
+type UserFacebook struct {
```
drsnyder: I would like to propose with start labeling these in a way that makes it clear that they are associated with data access. Something like `FacebookUserDAO`.

controllers/facebook_controller.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/Wikia/helios/models"
+	"github.com/Wikia/helios/service"
+	"github.com/influxdb/influxdb/client"
+)
+
+type FacebookController struct {
+	facebookConfig  *config.FacebookConfig
+	facebookService *service.FacebookService
+	influxdbClient  *client.Client
+	userStorage     *models.UserStorage
```
drsnyder: Minor: if possible, use the [interface](https://github.com/Wikia/helios/blob/master/models/user_storage.go#L16).

controllers/facebook_controller.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func NewFacebookController(influxdbClient *client.Client,
+	storageFactory *models.StorageFactory,
+	facebookConfig *config.FacebookConfig) *FacebookController {
+
+	controller := new(FacebookController)
+	controller.influxdbClient = influxdbClient
+	controller.userStorage = storageFactory.GetUserStorage(models.MainStorage)
```
drsnyder: Are we going to [shadow this](https://github.com/Wikia/helios/blob/master/controllers/users_controller.go#L50) or cut-over all at once?

controllers/facebook_controller.go
```diff
@@ -0,0 +1,110 @@
...
+			logger.GetLogger().ErrorErr(err)
+			fmt.Fprintf(w, "parse error")
+			return
+		}
+
+		userFbMap := models.NewUserFacebookMap(fbUserId, userId, appId)
+		controller.createUserFacebookMap(w, userFbMap)
+	} else {
+		fmt.Fprintf(w, "exist")
+	}
```
drsnyder: There are multiple levels of abstraction being used here. I think all of the fb and model interaction should happen in the service. The controller method should be limited to message passing from the client, to the service, and back to the client.

models/user_storage.go
```diff
@@ -173,6 +177,31 @@ func (userStorage *UserStorage) FindByName(userName string, mustExist bool) (*Us
 	return user, nil
 }
 
+func (userStorage *UserStorage) CreateUserFacebookMap(userFbMap *UserFacebookMap) error {
```
drsnyder: Can this be `SaveFacebookMap()` so it's consistent with the existing interface? Please also add it to the `UserDAO` interface.

models/user_storage.go
```diff
@@ -173,6 +177,31 @@ func (userStorage *UserStorage) FindByName(userName string, mustExist bool) (*Us
...
+	err := userStorage.masterMap.Insert(userFbMap)
+	if err != nil {
+		logger.GetLogger().ErrorErr(err)
+		return err
+	}
+
+	return nil
+}
+
+func (userStorage *UserStorage) FindByFbId(fbId int64) (*UserFacebookMap, error) {
```
drsnyder: Same here. Can we rename it so that it's consistent with `UserDAO` and add it to the interface. I chose `Lookup` arbitrarily but I had to use a different name because the interfaces were changing.

controllers/facebook_controller.go
```diff
@@ -0,0 +1,231 @@
...
+	if r.ParseForm() != nil {
+		setRequestError(resp, "parse error", "request", "request_parse_error")
+		return
+	}
+
+	userId, errors := controller.facebookService.LookupWikiaUserId(r.Form.Get(FB_ACCESS_TOKEN))
+
+	if errors.IsEmpty() && userId == 0 {
+		errors.Add(MissingFbMap)
+	}
```
drsnyder: This is logic that should be in the service and tested there. The way our controllers are current we can't test for this.

ArturKlajnerok: @drsnyder I know, I'm in the middle of moving all logic from tokenHandler to facebook service, but I need to introduce few changes first.

nmonterroso: It depends on who's asking for the `WikiaId` - in the case of login, it's an error if the user doesn't exist, but in the case of registration, it's an error if the user does exist.

nmonterroso: Also, `facebook_controller` is testable now, and we do test this case https://github.com/Wikia/helios/pull/42/files#diff-b7d5567daad5d6d0590544f9c44efe35R58

controllers/facebook_controller.go
```diff
@@ -0,0 +1,188 @@
...
+
+type FacebookController struct {
+	server          *osin.Server
+	deferror        *osin.DefaultErrors
+	clientConfig    *config.DefaultClientConfig
+	facebookService service.FacebookService
+	influxdbClient  *client.Client
+	userStorage     *models.UserStorage
+	// TODO: oauth_controller shares this functionality
+	allowMultipleAccessTokens bool
```
drsnyder: This is no longer an issue right? Can it be removed?

service/facebook.go
```diff
@@ -0,0 +1,117 @@
...
+
+	userFacebookMap, err := facebookService.userFacebookMapDAO.LookupByFbId(userData.UserId())
+	if err != nil {
+		errors.Add(common.GenericDataAccessError)
+		return nil, errors
+	}
+
+	if userFacebookMap != nil {
+		fbConnectedUser.wikiaUserId = userFacebookMap.UserId
+		fbConnectedUser.connected = true
```
drsnyder: As a general practice I don't think we should set private members on a struct especially one that has the `*Value` suffix. Also, I wonder if the `connected` logic is necessary. if the `facebookService.userFacebookMapDAO.LookupByFbId(userData.UserId())` call returns a `nil` user, that is sufficient enough to signal that the user is not connected.

This block could then become:

```go
userFacebookMap, err := facebookService.userFacebookMapDAO.LookupByFbId(userData.UserId())
if err != nil {
  errors.Add(common.GenericDataAccessError)
}

if userFacebookMap != nil {
   return NewFacebookConnectedUser(userData.UserId(), userData.AppId(), userFacebookMap.UserId), errors
}

return nil, errors
```

models/user_storage.go
```diff
@@ -57,14 +66,15 @@ func NewUserStorage(table string, dbConfig *config.DbConfig) *UserStorage {
 	return userStorage
 }
 
-func (userStorage *UserStorage) Save(user domain.User) (bool, error) {
-	err := userStorage.masterMap.Insert(FromDomain(user))
+func (userStorage *UserStorage) Save(user domain.User) (bool, int64, error) {
```
drsnyder: I think if we are going to return the user id as a third return value we might as well return `(models.User, error)`. This is equivalent to the too many parameters code smell. The `(nil, error)` case represents a failure.

If we return `(nil, error)` it also eliminates the need for `0` here which is ambiguous. We have records in our databases where the user id is 0. I think this will come back to haunt us later.

drsnyder: We are introducing `Save` in this branch so I think now is the time to do it right and decide on the interface we want to support.

service/facebook.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if fbConnectedUser == nil || errors != nil {
+		errors.Add(common.GenericDataAccessError)
+		return false, errors
+	}
+
+	if fbConnectedUser.IsConnected() {
+		errors.Add(ExistingFbMapError)
+		return false, errors
+	}
```
drsnyder: If we use nil instead of `IsConnected` then this could become:

```go
if fbConnectedUser != nil {
    errors.Add(ExistingFbMapError)
    return false, errors
}

if errors != nil {
   errors.Add(common.GenericDataAccessError)
   return false, errors
}

created, userValidationErrors := facebookService.userService.CreateUser(wikiaUser)
```

service/facebook.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if created {
+		userFbMap := models.NewUserFacebookMap(fbConnectedUser.FacebookAppId(), userId, fbConnectedUser.FacebookAppId())
+		saved, err := facebookService.userFacebookMapDAO.SaveUserFacebookMap(userFbMap)
+		if err != nil {
+			return false, userValidationErrors
+		}
+		return saved, nil
+	}
+
```
drsnyder: Should this be wrapped in a transaction? This is starting to get more complicated... Come to think of it, I think we should decouple the user creation and the anti spoofing update and the welcome email.

ArturKlajnerok: Yeah, this is moved from controller in this form, but I'm working on changing it to transaction :+1:

controllers/facebook_controller.go
nmonterroso: Nah, this is still needed. [SERVICES-362](https://wikia-inc.atlassian.net/browse/SERVICES-362)

models/user_storage.go
ArturKlajnerok: Fixed to return `(*User, error)` in : 
https://github.com/Wikia/helios/commit/d4a5733f657ba81e944dd7c6739fbce58f833453

service/facebook.go
ArturKlajnerok: We now use `LookupFacebookConnectedUser` in two cases, one is when we check is mapping exist to authenticate user in `/facebook/token` and get the wikia user id to create a access token.
Second is `ConnectFacebookUser` where want to be sure that mapping don't exist but we still want to get facebook user id and app id to create the mapping. That's way I introduced `connected` flag. 

As for "private members on a struct" do you suggest that we should make them exported ?

service/facebook.go
ArturKlajnerok: As on the upper comment, we need the `IsConnected` flag as even if user is not connected we want to grab some params to create the connection.

service/facebook.go
drsnyder: In hindsight I don't think I thought this through enough when I suggested this interface. Looking at it now I think `LookupFacebookConnectedUser` should be decoupled from verifying the facebook token. In it's current form I think it's doing too much.

As a caller of `LookupFacebookConnectedUser` I would expect it to just lookup the user. As such I would expect to either get `nil` or a valid user.

> As for "private members on a struct" do you suggest that we should make them exported ?

No. I think we should favor immutable objects over mutable ones. That is, mutable objects should be the _exception_ and not the rule. 

In this case the notion that a user is "facebook connected" shouldn't change as the program executes. Either the user is facebook connected or not and the existence of a facebook connected user should be sufficient to represent that state. Providing the ability to change that state is opening the program up to subtle bugs.

I don't think we have to change this before you merge but it is something that we should come back to and refactor. Probably in this sprint with SERVICES-337.

nmonterroso:
LGTM :ship:

service/facebook.go
```diff
@@ -0,0 +1,119 @@
...
+func (user *FacebookConnectedUserValue) WikiaUserId() int64 {
+	return user.wikiaUserId
+}
+
+func (user *FacebookConnectedUserValue) IsConnected() bool {
+	return user.connected
+}
+
+func (user *FacebookConnectedUserValue) Exists() bool {
+	return user != nil
```
drsnyder: This seems tenuous. The user exists and is connected if fbappid, fbuserid, and wikiauserid are all non nil integers.

drsnyder:
I just had the one comment about `user != nil`. Other than that, LGTM.

ArturKlajnerok:
@drsnyder I've removed the interface do we don't need to externalize that nil check anymore.

PR #43, opened 2015-05-28
SERVICES-338 /register endpoint -> /users resource

drsnyder:
Change the `/register` endpoint to the `/users` resource. As part of this the terminology around "register" was also changed to "create" for more internal consistency.

Note that this *should* be 100% backwards compatible with existing clients and configuration. `gcfg` does not allow unrecognized configuration values to be added but does allow them to be missing. In the case of a missing configuration variable the [zero value](https://golang.org/ref/spec#The_zero_value) is used which in the case of booleans is `false`. To confirm, I tested master against the server built on this branch.

https://wikia-inc.atlassian.net/browse/SERVICES-338

TODO:
 * [ ] chef changes to add `disable-user-creation`
 * [x] MW changes to switch `/register` to `/users`


controllers/user_controller.go
```diff
@@ -50,30 +50,40 @@ func NewRegisterController(
...
+	// 2. add disable-user-creation and remove disable-registration from chef
+	// 3. remove DisableRegistration from config/config.go
+	controller.disableUserCreation = serverConfig.DisableRegistration
 	controller.userValidator = validation.NewUserValidator(validation.AllValidationFuns...)
 
+	// FIXME: remove once MW has been transitioned to use the /users resource.
+	// This is for backwards compatibility with existing clients
 	http.HandleFunc("/register",
-		timedOriginHandler(controller, "registerHandler", controller.registerHandler))
+		timedOriginHandler(controller, "registerHandler", controller.usersHandler))
```
nmonterroso: should this be changed to `userHandler`?

controllers/user_controller.go
```diff
@@ -50,30 +50,40 @@ func NewRegisterController(
...
+
+	http.HandleFunc("/users",
+		timedOriginHandler(controller, "usersHandler", controller.usersHandler))
+
 	http.HandleFunc("/username/validation",
```
nmonterroso: I wonder if this now makes more sense as `users/validate/name` now?

nmonterroso:
Should we (at some point) change the login resource as well?

nmonterroso:
minor comments otherwise LGTM

controllers/user_controller.go
drsnyder: I think I want to leave that as `registerHandler`. That string is used for metrics so this will help us determine if we can safely remove it.

controllers/user_controller.go
drsnyder: That URI will go away after https://wikia-inc.atlassian.net/browse/SERVICES-342.

controllers/user_controller.go
nmonterroso: ah, ok

controllers/user_controller.go
nmonterroso: Is that the plan? It seems weird that `dryrun` would be used for only username validation. Consider that only a valid username is sent to `dryrun`, then it would respond with response code `400 - missing email, password, birthdate`. 

As a developer I would expect that I could rely on a `200` telling me that the username is fine, and a non-200 would tell me I need to look at the error list to find username errors and adjust my ui accordingly. But if we only send a username to `dryrun` it will always respond with a `400`, unless we plan to have a way to tell `dryrun` we're only validating a few fields, in which case ignore me :P

drsnyder: Making HTTP requests to `/username/validation` is very RPC-like. That's not necessarily a bad thing in that it can be pragmatic, but the semantics of the HTTP request start to become very custom.

For example, what does it mean to `POST` (verb) to `/username/validate` (also a verb)? Just because you get a 200 back doesn't mean that you can create the account which is what you wanted to do. Also what happens when we also want to validate the birthday? Do we need another URI for that?

Using `users/?dryrun` (though closer to REST) still has some drawbacks. For example, just because the `POST` responds with `200` doesn't mean you can create the account. Someone may have already taken the user name between the dry-run and the followup `POST`. Ideally we would reserve the username for some fixed period of time.

I suppose one pragmatic adjustment would be to provide a query parameter to provide guidance to the server as to what you want validated. For example, you might want to `POST` `/users?dryrun&field=username`. Thoughts?

drsnyder:
> Should we (at some point) change the login resource as well?

 Do you mean `/token`?

nmonterroso:
Yah I guess that doesn't make sense considering we are going to oauth at some point, and `/token` probably makes more sense in that context?

controllers/user_controller.go
nmonterroso: Yah, I'm not suggesting that `/username/validate` is the best endpoint for it, I'm just saying that if we're going to use `dryrun` to authenticate only a username, I'd expect that a `200` means I can show the green check mark next to the username, and a `400` means that I can't. But in this case, if we don't pass along "only check the username" somehow to `dryrun`, I'd get a `400`, then I'd have to iterate through the errors to check if there's an error with the username.

Of course, if we can say `/users/?dryrun&fields=username` then that's probably best, then we could use it with `email`, `password`, or any combination of them to validate only those fields, but the `dryrun` story does not specify that, and it implies that it will check against all fields.

drsnyder: I updated the ticket with the gist of this thread.

drsnyder:
> Yah I guess that doesn't make sense considering we are going to oauth at some point, and /token probably makes more sense in that context?

I haven't put much thought into it but having a noun on the URI as apposed to a verb on the surface seems more what you would expect.

controllers/user_controller.go
ArturKlajnerok: @nmonterroso passing here `registerHandler` is used only for performance monitoring

ArturKlajnerok:
Lgtm :+1:

PR #44, opened 2015-05-29
Request for comment: creating a user service

drsnyder:
This service will provide a facade in front of the database layer and provide a container within which to place the user creation logic so it can be moved out of the registration controller and shared between the
standard user creation and the facebook user creation controllers.

This was something that @nmonterroso, @ArturKlajnerok, and I talked about today regarding the need for an additional abstraction for user creation/lookup.

/cc @Wikia/services-team


domain/facebookuser.go
```diff
@@ -0,0 +1,15 @@
+package domain
+
+type FacebookUser interface {
```
nmonterroso: should `FacebookUserValue` implement this interface?

nmonterroso:
This is great, though `UserService` probably needs a few more functions - `usernameExists(name string)`, `emailExists(name string)`, `fbIdExists(fbId int64)` and probably the other functions that are currently exposed by `UserStorage`.

domain/facebookuser.go
drsnyder: Yeah, I hadn't gotten that far by the time I had to stop yesterday. I wanted to share the prototype so we could get started on it today.

drsnyder:
@Wikia/services-team  I think this is ready for review. I can't quite replace the user creation logic in the controllers until I port over the external user validation that validates users via MW and phalanx.

ArturKlajnerok:
@drsnyder Look solid to me, I have only one question about `FacebookUser` as we won't need to create any specific user for facebook and only mapping of wikia user id and facebook user id.

service/user_test.go
```diff
@@ -0,0 +1,115 @@
...
+type UserServiceTestSuite struct {
+	suite.Suite
+	userDAO   *MockedUserDAO
+	validator *MockedValidator
+	user      *domain.UserValue
+}
+
+func (suite *UserServiceTestSuite) SetupTest() {
+	suite.userDAO = new(MockedUserDAO)
+	suite.user = domain.NewUser("bill", "password", "bill@wikia-inc.com", "1990-05-01")
```
ArturKlajnerok: Values are repeated, we could move them to const.

drsnyder:
> I have only one question about FacebookUser as we won't need to create any specific user for facebook and only mapping of wikia user id and facebook user id.

@ArturKlajnerok ok, let's talk about how this association should be represented within the domain. Something like this might make sense:

```go
facebookService.LinkUser(domain.FacebookUser)
```

What this would do is link the facebookuser id to the domain user using the username.

service/user_test.go
drsnyder: Ok, I moved these to constants in 2864644.

nmonterroso:
LGTM

PR #46, opened 2015-06-01
Use the DisableUserCreation configuration and remove DisableRegistration

drsnyder:
This is the next step in the config dance to change registration to users.

https://wikia-inc.atlassian.net/browse/SERVICES-338

 1. done: deploy helios with the new structure
 2. make the changes in chef to add the new config
 3. ---> we are here: remove the old config from helios and use the new one (it did not exist in the config file in chef so we couldn't use it)
 4. remove the old config from chef
 5. remove the old config variable from helios


ArturKlajnerok:
Lgtm :+1:

PR #50, opened 2015-06-01
move phalanx/spoof validation checks to validation package

nmonterroso:
@drsnyder @ArturKlajnerok 
https://wikia-inc.atlassian.net/browse/SERVICES-355

make phalanx and anti spoof checks compliant with `IsValidFunc` interface

todo:
- [x] - move mocks to `validation/mocks.go`

PR #51, opened 2015-06-01
Validation cleanup

nmonterroso:
@drsnyder @ArturKlajnerok 
https://wikia-inc.atlassian.net/browse/SERVICES-355

Adding the validators for username/email existing after the changes in https://github.com/Wikia/helios/pull/50. Also removing some deprecated code.


controllers/users_controller.go
```diff
@@ -105,19 +111,17 @@ func (controller *UsersController) usersHandler(w http.ResponseWriter, r *http.R
...
-		modelUser := models.FromDomain(user)
-
-		if controller.createUser(resp, modelUser) {
-			controller.onUserCreationComplete(modelUser, controller.getLangCodeFromRequest(r))
-		}
 	}
+
+	modelUser := models.FromDomain(user)
+	controller.onUserCreationComplete(modelUser, controller.getLangCodeFromRequest(r))
+	controller.setOutputSuccess(resp, "User created")
```
drsnyder: Should this be in an `else` block? Also, I think we are missing the case where `! userValidationErrors.IsEmpty()`.

models/user_storage.go
```diff
@@ -109,55 +109,12 @@ func (userStorage *UserStorage) LookupByEmail(email string) (*User, error) {
...
-		return false, err
-	}
-
-	return count > 0, nil
-}
-
-// Deprecated
 func (userStorage *UserStorage) FindByName(userName string, mustExist bool) (*User, error) {
 	user := new(User)
 	query := fmt.Sprintf("select * from %s where user_name=?", userStorage.tableName)
```
drsnyder: I think this one can go away too.

drsnyder:
Net negative!

controllers/users_controller.go
nmonterroso: Yah, the tests noticed this for me, and i fixed it in 400e6f5.

models/user_storage.go
nmonterroso: I wish. It's being used in [oauth_controller](https://github.com/Wikia/helios/blob/master/controllers/oauth_controller.go#L98)

service/user.go
```diff
@@ -45,7 +46,7 @@ func (service *UserServiceValue) CreateUser(user domain.User) (bool, *errorlist.
 
 	existingEmail, err := service.userDAO.LookupByEmail(user.Email())
 	if err != nil {
-		return false, nil
+		return false, validationErrors
```
drsnyder: :+1: better than `nil`

controllers/users_controller.go
```diff
@@ -105,18 +111,16 @@ func (controller *UsersController) usersHandler(w http.ResponseWriter, r *http.R
 	}
 
 	user := createDomainUserFromForm(r)
-	isDomainObjectValid, userValidationErrors := controller.userValidator.Validate(user)
+	created, userValidationErrors := controller.userService.CreateUser(user)
 
-	if userValidationErrors.ContainsError(validation.GenericDataAccessError) {
+	if created {
+		modelUser := models.FromDomain(user)
+		controller.onUserCreationComplete(modelUser, controller.getLangCodeFromRequest(r))
```
drsnyder: In hindsight, we should have moved this to the service.

PR #52, opened 2015-06-02
Remove disable-registration from the config

drsnyder:
Remove the `disable-registration` configuration parameter from the helios config.

This is the last step in https://wikia-inc.atlassian.net/browse/SERVICES-338.


ArturKlajnerok:
Lgtm :+1:

PR #53, opened 2015-06-02
Move the facebook API to gateway & create an interface for testing and injection

drsnyder:
This creates a `FacebookGateway` interface that provides `ValidateToken(accessToken string) (AuthenticatedUser, error)`. This should be the only method we need to either connect a facebook user or log one in.

Merging this will break the current branch because the service changes. The fixes should be straightforward though. I'm fine making those fixes here or after the merge.


ArturKlajnerok:
Lgtm :+1:

drsnyder:
I also pushed the gateway changes up into the controller.

PR #54, opened 2015-06-04
RFC: Consolidate user creation post processing & move the validation into the domain package

drsnyder:
We have some common functionality that maps functions over `domain.User`. The first is the validation, the second is the user post-processing. The post processing currently includes emailing notifications and updating the anti-spoofing.

While thinking a lot about the later (post processing) I noticed it looked like the former. I also realized that a lot of the validation code and logic was very specific to `domain.User`. So, I moved everything from the `validation` package into `domain`.

I also made the validation interfaces more generic. The hope is that in the user service we can:

```go
go service.postCreateCommands.Execute(user)
```

And that will run all of the post processing command required. Benefits (I think):
 * We only have to add one dependency to the user service (or none, we could just call it inline)
 * This functionality can be easily extended to add more commands
 * We don't have to pass in the `controller.disableUserCreation` to the service
 * We can reuse existing, tested functionality

Comments? Objections? Concerns?

https://wikia-inc.atlassian.net/browse/SERVICES-360


ArturKlajnerok:
Lgtm :+1:

drsnyder:
@nmonterroso, @ArturKlajnerok sorry for how bloated this PR looks. Git is recording the moving of the validation related files into domain as deletion and addition respectively. The meat of this is in the user service and the users_controller and the controller helpers.

Instead of creating a new structure for the wikia and phalanx clients, I created a helper that creates the command functions and closes over the clients and the language codes. I'm then passing in the object to the `CreateUser` method to be called after the user is created. So far, these are all one line functions that just return true/false and the error if any. I guess if it gets more complicated we can move it out and test it.

controllers/helpers.go
```diff
@@ -32,3 +34,18 @@ func setOutputProblem(resp *osin.Response, p problem.Problematic, problemTitle s
 	resp.Output = p.AsProblem(problemTitle).AsMap()
 	resp.StatusCode = statusCode
 }
+
+func createPostCreationCommandList(service service.UserService, langCode string, disableUserCreation bool) domain.UserCommandList {
```
nmonterroso: Will this function's parameters get too unwieldy at some point? I wonder if we should create a function interface like we did for user validation?

nmonterroso: Oh, nevermind I should have looked two lines down. Looks like we do have that interface, and this is just a shortcut to create the list. I still think this function might have too many parameters in the not-too-distant future. We'd be adding to this next sprint already.

nmonterroso:
LGTM (but tests are busted :()

drsnyder:
> LGTM (but tests are busted :()

If you can get it to pass once :shipit:

controllers/helpers.go
drsnyder: If we need another command (like creating the user pref) we can just add to the list of commands in `service.PostCreationCommandListFactory`.

PR #55, opened 2015-06-08
Be more defensive in referencing the error pointer in Add

drsnyder:
If `Add` is given a null pointer it will crash the process.

/cc @ArturKlajnerok @nmonterroso


nmonterroso:
I wonder if we should log when we try to add a `nil` error, so that we can catch these things and make it a point to insert a valid `error`, instead of silently ignoring

ArturKlajnerok:
+1 for logging, other then that lgtm :+1:

PR #56, opened 2015-06-09
User storage return model user instead bool

sqreek:
Provide returned model.User to get User id for sending it to user preferences service


ArturKlajnerok:
This is already done in : https://github.com/Wikia/helios/pull/42

PR #57, opened 2015-06-10
Add default client id/secret to /token request

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-361

Adds `grant_type="password"` and `client_id`/`client_secret` from config to incoming `/token` requests. This allows us to send over just `username` and `password` and receive a valid token


ArturKlajnerok:
Please updated the swagger annotations on `tokenHandlerPassword` as few params are now not required.

nmonterroso:
good point, @ArturKlajnerok, added in c3eae5b

controllers/oauth_controller.go
```diff
@@ -177,6 +185,17 @@ func (controller *OAuthController) setOutputError(resp *osin.Response, errorCode
 	resp.Output["error_description"] = controller.deferror.Get(errorCode)
 }
 
+func addPasswordGrantValues(config *config.DefaultClientConfig, form *url.Values) {
+	if form.Get(helpers.GrantTypeKey) == "" {
+		form.Set(helpers.GrantTypeKey, osin.PASSWORD)
+	}
+
+	if form.Get(helpers.ClientIdKey) == "" || form.Get(helpers.ClientSecretKey) == "" {
```
ArturKlajnerok: Maybe we should split that, as it can be weird for client that if won't provide one of the params both are set to default ?

ArturKlajnerok:
Lgtm :+1:

PR #59, opened 2015-06-11
Updated oauth docs and changed docs order

ArturKlajnerok:
Latest version based on changes in oauth controller and updated resource order.


nmonterroso:
LGTM

PR #60, opened 2015-06-11
[SERVICES-401] Connection accounts in facebook service

ArturKlajnerok:
This will provide proper method to connect facebook and wikia accounts via interface : ```ConnectUsers(userId, appId int64, accessToken string) (bool, *errorlist.ErrorList) ```

Facebook service will be refactored to provide clean way to : 
- [x] Check if facebook and wikia accounts are connected
- [x] Create wikia account and connect it with facebook account
- [x] Connect facebook and wikia accounts
- [x] Check if account connection already exist

Request for feedback :smile_cat: 
/cc @Wikia/services-team


service/facebook.go
```diff
@@ -58,50 +74,33 @@ func (facebookService *FacebookServiceValue) ConnectFacebookUser(accessToken str
...
-
-func (user *FacebookConnectedUser) FacebookUserId() int64 {
-	return user.facebookUserId
-}
+func (facebookService *FacebookServiceValue) isUserConnected(fbUserId, fbAppId int64) (bool, *errorlist.ErrorList) {
+	errors := errorlist.NewErrorList()
+	userFacebookMap, err := facebookService.userFacebookMapDAO.LookupUserFacebookMap(fbAppId, fbUserId)
+	if err != nil {
+		errors.Add(errorlist.GenericDataAccessError)
+		return true, errors
```
nmonterroso: Should these be reversed? Seems odd for a function named `isUserConnected` to return `true` when an `error` is seem

service/facebook.go
```diff
@@ -31,25 +31,41 @@ func NewFacebookService(userFacebookMapDAO models.UserFacebookMapDAO, facebookGa
...
-	fbConnectedUser, errors := facebookService.LookupFacebookConnectedUser(accessToken)
+func (facebookService *FacebookServiceValue) ConnectUsers(userId, appId int64, accessToken string) (bool, *errorlist.ErrorList) {
+	facebookUser, errors := facebookService.facebookGateway.ValidateToken(accessToken)
+	if facebookUser == nil || !errors.IsEmpty() {
+		return false, errors
+	}
 
-	if fbConnectedUser == nil {
+	if connected, errors := facebookService.isUserConnected(facebookUser.UserId(), facebookUser.AppId()); connected {
 		return false, errors
```
drsnyder: Should this add an error to `errors` that indicates that the user was already mapped?

service/facebook.go
```diff
@@ -31,25 +31,41 @@ func NewFacebookService(userFacebookMapDAO models.UserFacebookMapDAO, facebookGa
...
+}
+
+func (facebookService *FacebookServiceValue) CreateAndConnectUser(accessToken string, wikiaUser domain.User) (bool, *errorlist.ErrorList) {
+	facebookUser, errors := facebookService.facebookGateway.ValidateToken(accessToken)
+	if facebookUser == nil || !errors.IsEmpty() {
+		return false, errors
+	}
+
+	if connected, errors := facebookService.isUserConnected(facebookUser.UserId(), facebookUser.AppId()); connected {
 		return false, errors
```
drsnyder: Same as the comment above about the error.

service/facebook.go
```diff
@@ -31,25 +31,41 @@ func NewFacebookService(userFacebookMapDAO models.UserFacebookMapDAO, facebookGa
...
 
-	if fbConnectedUser == nil {
+	if connected, errors := facebookService.isUserConnected(facebookUser.UserId(), facebookUser.AppId()); connected {
 		return false, errors
 	}
 
-	if fbConnectedUser.IsConnected() {
-		errors.Add(ExistingFbMapError)
+	userFacebookMap := models.NewUserFacebookMap(facebookUser.UserId(), userId, facebookUser.AppId())
+	saved, err := facebookService.userFacebookMapDAO.SaveUserFacebookMap(userFacebookMap)
```
drsnyder: I think it would be nice to move the `models.NewUserFacebookMap` call into `SaveUserFacebookMap`. That would remove at least 2 lines from this file. It also seems more appropriate for the model method to be concerned about creating the structure to save.

service/facebook.go
ArturKlajnerok: Made it more explicit in : 213120b

service/facebook.go
ArturKlajnerok: Errors are added inside of `isUserAlreadyConnected` to reduce code duplication : https://github.com/Wikia/helios/pull/60/files#diff-08878cd99cbc2fc48f2bc410e2328d38R92

service/facebook.go
ArturKlajnerok: Errors are added inside of `isUserAlreadyConnected` to reduce code duplication : https://github.com/Wikia/helios/pull/60/files#diff-08878cd99cbc2fc48f2bc410e2328d38R92

service/facebook.go
ArturKlajnerok: Moved in : d217b0c

nmonterroso:
LGTM - anything else from your side @drsnyder

PR #61, opened 2015-06-12
Work-in-progress: Straw man for the facebook connect controller

drsnyder:
This PR creates a straw man facebook connect controller.  I added the gorrilla toolkit mux package to handle parameterized URIs so we don't have to parse URIs ourselves.

https://wikia-inc.atlassian.net/browse/SERVICES-402

/cc @Wikia/services-team 

TODO:
 * [x] get the user id from the URI
 * [x] get the user id from the X-Wikia-UserId header
 * [x] get the app id from the URI
 * [x] get the fb_access_token from the body
 * [x] call facebookService.ConnectUsers(userid, authUserId, fbAppId, fbAccessToken
 * [x] handle errors if there are any
 * [x] return 200 otw


controllers/helpers.go
```diff
@@ -20,11 +22,31 @@ const (
...
+	vars := mux.Vars(req)
+	h(w, req, vars)
+}
+
 type TimedController interface {
 	GetMetricsClient() *client.Client
 }
 
-func timedOriginHandler(controller TimedController, handlerName string, fn func(http.ResponseWriter, *http.Request)) http.HandlerFunc {
+func timedOriginMuxHandler(controller TimedController, handlerName string, fn MuxVarsHandler) http.HandlerFunc {
```
nmonterroso: This looks very similar to the implementation of `timedOriginHandler`, is there any way to consolidate these?

nmonterroso:
LGTM

controllers/helpers.go
drsnyder: Yes! Thanks to the fact that a function interface can be a receiver! See 57ee959.

nmonterroso: Are you a wizard?

drsnyder: No. I'm working on [magician](https://www.linkedin.com/in/drsnyder) but I don't think I'm there yet.

PR #62, opened 2015-06-15
add user preference service request while registration.

sqreek:
added User preference service request while registration
/cc @Wikia/services-team


sqreek:
Question: From where I should take tag for user-preference service?
Already implemented in config, but can be also from environment, or consul.

config/config.go
```diff
@@ -63,6 +67,7 @@ type Config struct {
 	RedisMaster  RedisInstanceConfig `gcfg:"redis-master"`
 	RedisSlave   RedisInstanceConfig `gcfg:"redis-slave"`
 	Client       DefaultClientConfig `gcfg:"default-client"`
+	UserPreference UserPreferenceServiceConfig `gcfg:"user-preference-service"`
```
nmonterroso: can we `gofmt` this?

nmonterroso:
@sqreek I think we might want to move to configs fetched from Consul at some point, but for now we should keep consistent with other configs (in YAML)

controllers/users_controller.go
```diff
@@ -112,6 +115,7 @@ func (controller *UsersController) usersHandler(w http.ResponseWriter, r *http.R
 	userData, userValidationErrors := controller.userService.CreateUser(user)
 
 	if userData != nil {
+		controller.updateUserPreference(r, userData)
```
nmonterroso: Instead of calling this directly we should leverage the `ApplyToUserFunc` interface, move `updateUserPreferences` to the `UserServiceValue` struct, and add `updateUserPreferences` to `UserServiceValue.PostCreationCommandListFactory` so we can keep all of our user post-creation steps in one place.

service/user_preference.go
```diff
@@ -0,0 +1,101 @@
...
+	url := fmt.Sprintf("%s/%d", ipAddress, userId)
+
+
+
+	req, err := http.NewRequest("PUT", url, bytes.NewBuffer(jsonBytes))
+	if err!=nil {
+		return err
+	}
+
+	req.Header.Set("user_id", strconv.FormatInt(userId, 10))
```
nmonterroso: I don't think this header is used by `user-preference`

service/user_preference.go
```diff
@@ -0,0 +1,101 @@
...
+	ipAddress, err := service.consulResolver.ResolveURI("user-preference", service.userPreferenceTag)
+	if err!=nil {
+		return err
+	}
+
+	url := fmt.Sprintf("%s/%d", ipAddress, userId)
+
+
+
+	req, err := http.NewRequest("PUT", url, bytes.NewBuffer(jsonBytes))
```
nmonterroso: We've had issues before with `http.Request` objects not properly closing before, causing issues with subsequent requests to fail, so we might need to add `req.Close = true`. see https://github.com/Wikia/go-commons/pull/9/files

config/config.go
sqreek: I was used intellij for that, and it seems to not work correctly :(

service/user_preference.go
sqreek: it's used in authentication, and it is compared [here](https://github.com/Wikia/pandora/blob/4dab65e65d96b7c00d3e870277b5a698fee10813/service/user-preference/src/main/java/com/wikia/userpreference/resources/UserPreferenceResource.java#L73)

service/user_preference.go
sqreek: added req.Close = true for handle it

sqreek:
@nmonterroso , but where in helios we have config files in yaml?
Maybe I should rephrase question:
where Helios should look for tag of user-preference service that it need to use?

controllers/users_controller.go
sqreek: I thougth about it, but:
ApplyUserFunc need to use domain.User, and userPreferences need additional userId (so model.User). And also minor it will need to extend it for user preferences field.
Second case is easy, and doesn't affect flow, but changing it from domain.User to model.User is bigger refactor, and mix scopes of User's (model and domain)

I will leave it for now as it is.

service/user_preference.go
nmonterroso: Does it? For authentication my understanding is that we use `X-Wikia-UserId`, which is set below. In user-preference the `userId` is taken from the path.

controllers/users_controller.go
nmonterroso: I don't know if we need to change it to use `model.User`, I think we can change it to also supply the `userId` as an additional parameter, so the interface would be 
```go
type ApplyToUserFunc func(userId int, user domain.User) (bool, *errorlist.KeyedError)
```

@drsnyder - thoughts?

service/user_preference.go
sqreek: ou ... yes ... double checked, and you are right

controllers/users_controller.go
sqreek: and also we need to add user_preferences to domain.User or add another parameter, and refactor all existing ApplyToUserFunc.
If we decide to refactor this way, I want to do this in separate PR.

service/user_preference.go
```diff
@@ -0,0 +1,103 @@
...
+	if err!=nil {
+		return err
+	}
+
+	jsonBytes, err := json.Marshal(preferences)
+	if err!=nil {
+		return err
+	}
+
+	ipAddress, err := service.consulResolver.ResolveURI("user-preference", service.userPreferenceTag)
```
drsnyder: `url` would be a better variable name than `ipAddress`. `ResolveURI` returns a string of the form `http://...`.

service/user_preference.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	url := fmt.Sprintf("%s/%d", ipAddress, userId)
+
+	req, err := http.NewRequest("PUT", url, bytes.NewBuffer(jsonBytes))
+	if err!=nil {
+		return err
+	}
+
+	req.Header.Set("X-Wikia-UserId", strconv.FormatInt(userId, 10))
```
drsnyder: You can use `common.GatewayUserIdHeader` for `X-Wikia-UserId`.

service/user_preference.go
```diff
@@ -0,0 +1,103 @@
...
+	service.consulResolver = consul.DefaultResolver()
+	// prevent from calling random service with any tag
+	if userPreferenceTag =="" {
+		userPreferenceTag = "testing"
+	}
+	service.userPreferenceTag = userPreferenceTag
+	return service
+}
+
+func (service *UserPreferenceServiceValue) UpdateUserPreferences(userId int64, userPreferencesJSON string) error {
```
drsnyder: I think it would be nice to have this return `(bool, error)`. I think that's more idiomatic in cases like this where eternal IO is being performed.

drsnyder: Even better would be `(bool, *errorlist.KeyedError)`. Here is an [example](https://github.com/Wikia/helios/blob/master/service/user.go#L85).

service/user_preference.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if resp.StatusCode != 200 {
+		var userPreferenceResponse UserPreferenceServiceResponse
+		bs, _ := ioutil.ReadAll(resp.Body)
+		_ = json.Unmarshal(bs, &userPreferenceResponse)
+		return errors.New(userPreferenceResponse.Title)
+	}
+
+	return nil
+}
```
drsnyder: The level of abstraction in this "service" is different from the other "services" provided within helios. I think this is better described as a "gateway". See the [facebook gateway](https://github.com/Wikia/helios/blob/master/gateway/facebook.go) or the [MediaWiki gateway](https://github.com/Wikia/helios/blob/master/wikia/wikia.go) which should probably be moved to `/gateway`.

We have been trying to create a "service" layer within helios that handles some of the "verbs" within the system that aren't necessarily concepts within the domain-- e.g. `CreateUser`, `FacebookConnectUser` etc. I think the functionality here be better suited to a "gateway" that is injected into the user service.

This functionality is very closely tied to `CreateUser`. @nmonterroso mentioned that it might be best to use `ApplyToUserFunc`. I think that's tempting but to me there is more cohesion with adding that to the logic of `CreateUser`. At some point we'll want to put a transaction boundary around creating the user and the preferences so those two will probably always happen along side each other. As a "responsibility" creating a user now also means initializing their preferences with those provided.

If you follow that rational then a user preference "gateway" should be injected into the user service (`service/user.go`) that satisfies the interface `UpdateUserPreferences(user models.User, preferences []UserPreference) (bool, *errorlist.KeyedError)`. You could then use the `models.User` returned from `userservice.CreateUser` to update the user preferences.

I think this will maintain the same level of abstraction in the helios "service" layer and increase the cohesion around the concept of creating a user. It should also provide a migration path to potentially combining the two in some form of transaction eventually.

I found [this article](http://gorodinski.com/blog/2012/04/14/services-in-domain-driven-design-ddd/) helpful in thinking about the relationship between a domain and a "service". I think we are in a transitionary phase with helios where the domain is currently anemic but evolving. It's important that we at least keep the levels of abstraction at or near the same level while we make this transition to piling on more functionality in helios.

nmonterroso:
@sqreek sorry, I guess I didn't finish my thought in that comment about the config yaml. There is currently no yaml in Helios. I was comparing it to having centralized configuration (like we do in Pandora, with YAML), but in Helios we have this from the config.ini, so I think the config.ini file is the right place.

service/user_preference.go
nmonterroso: Ah, good point about having it in the logic of `CreateUser`. Having the eventual transaction would cause us to move it there anyway, so I agree

sqreek:
ping @Wikia/services-team

controllers/helpers.go
```diff
@@ -50,11 +53,25 @@ func timedOriginHandler(controller TimedController, handlerName string, fn http.
 }
 
 func createDomainUserFromForm(request *http.Request) domain.User {
-	return domain.NewUser(
+
+	preferencesJSON := request.Form.Get("user_preferences")
```
drsnyder: Should this key be `preferences` instead of `user_preferences`? The context implies user.

domain/user.go
```diff
@@ -18,13 +18,15 @@ type User interface {
...
 type UserValue struct {
-	name     string
-	password string
-	email    string
-	birthday string
+	name            string
+	password        string
+	email           string
+	birthday        string
+	userPreferences []UserPreference
```
drsnyder: Minor: I think this could also be `preferences` since it's part of a "user". The accessor above could also be `Preferences()`.

gateway/user_preference_e2e_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+//func (suite *UserPreferenceServiceE2eTestSuite) TestUpdateUserPreferenceWithInvalidData() {
+//err := suite.service.UpdateUserPreferences(validUserId, `[{"not_valid_name":"date", "value":"1"}, {"name":"language", "value":"en"}]`)
+//assert.EqualError(suite.T(), err, ErrorNoUserPreferenceFound)
+//}
+
+//func (suite *UserPreferenceServiceE2eTestSuite) TestUpdateUserPreferenceWithInvalidJson() {
+//err := suite.service.UpdateUserPreferences(validUserId, `This is not a Json`)
+//assert.EqualError(suite.T(), err, ErrorInvalidJson)
+//}
```
drsnyder: Are these still valid assuming we change the second argument?

drsnyder:
Minor comments. Otherwise, LGTM

sqreek:
Last call before merge.

drsnyder:
:shipit:

PR #63, opened 2015-06-15
Removing duplicated user validation

ArturKlajnerok:
We are doing duplicated username and email validation in storage : 
https://github.com/Wikia/helios/blob/master/controllers/users_controller.go#L63
https://github.com/Wikia/helios/blob/master/models/user_dao_validator.go#L29
https://github.com/Wikia/helios/blob/master/service/user.go#L53

Removed it form user service, as now is done via validator.

/cc @Wikia/services-team


drsnyder:
:+1: thanks for doing that

nmonterroso:
LGTM

PR #64, opened 2015-06-16
WIP [SERVICES-408] Swagger docs api

ArturKlajnerok:
We need to expose swagger documentation via API to use it and test it with swagger-ui.

/cc @Wikia/services-team


controllers/swagger_controller.go
```diff
@@ -0,0 +1,48 @@
...
+	influxdbClient *client.Client
+}
+
+func NewSwaggerController(influxdbClient *client.Client) *SwaggerController {
+
+	controller := new(SwaggerController)
+	controller.influxdbClient = influxdbClient
+
+	rtr := mux.NewRouter()
+	rtr.HandleFunc("/swagger/",
```
nmonterroso: The rest of the services serve swagger at `/swagger.json`, is it possible to do this in Helios as well?

docs/docs.go
```diff
@@ -43,7 +43,7 @@ var resourceListingJson = `{
         "licenseUrl": "http://opensource.org/licenses/MIT"
     }
 }`
-var apiDescriptionsJson = map[string]string{"heartbeat": `{
+var ApiDescriptionsJson = map[string]string{"heartbeat": `{
     "apiVersion": "1.0.0",
     "swaggerVersion": "1.2",
     "basePath": "http://localhost:9500",
```
nmonterroso: I think the `basePath` needs to change, or even better somehow be dynamic. Also the `resourcePath` might also have to be dynamic, since in production it will be `/helios/facebook`, `/helios/users`, etc

nmonterroso:
I tried to convert the 1.2 swagger to 2.0, but it fails because the http method types are output as `httpMethod`, when the [1.2 swagger spec](https://github.com/swagger-api/swagger-spec/blob/master/versions/1.2.md) says to use `method`

controllers/swagger_controller.go
ArturKlajnerok: `/swagger/` this is default location and base path for swagger resources api in swagger 1.2

nmonterroso:
LGTM

PR #65, opened 2015-06-16
Valid Facebook Token (for non-valid app) returning 500

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-183

We found an issue where giving a valid facebook app token (but for an app that Helios doesn't recognize) was causing a 500-level response. This changes the response in that case to be a 400 response.

Also adding `gorilla` dependencies to `Godeps.json`


drsnyder:
lgtm

ArturKlajnerok:
Lgtm :+1:

PR #68, opened 2015-06-16
Move user-preference service to gateway then use the gateway in the user service

drsnyder:
This PR moves the user preference service functionality into the gateway and wires that functionality into the user service. Creating the user and adding the preferences go hand-in-hand so I think it makes sense for the functionality to be gathered in the user service.

https://wikia-inc.atlassian.net/browse/SERVICES-356

TODO:
 * [x] Create the user preference gateway
 * [ ] Wire the user preference gateway into the user service


drsnyder:
I didn't get as far as wiring it into the service. I just wired it into the controller in the same way the service was before.

errorlist/error_list_test.go
```diff
@@ -17,7 +16,7 @@ var (
 )
 
 func TestErrorList(t *testing.T) {
-	logger.InitLogger(common.AppName, logger.LogLevelDebug)
+	logger.InitLogger("foo", logger.LogLevelDebug)
```
drsnyder: This was to remove the cyclic dependency.

sqreek:
Thanks, I will take it from here

PR #69, opened 2015-06-19
[SERVICES-443] Cookie domain support

ArturKlajnerok:
We need to be able to configure domain that is used for setting the cookie, so we have control over supporting subdomains and different environments.

/cc @Wikia/services-team


drsnyder:
lgtm

PR #70, opened 2015-06-25
[SERVICES-408] Setting fixed basepath

ArturKlajnerok:
This should provide proper paths to resources from swagger docs api.

/cc @Wikia/services-team


nmonterroso:
lgtm

PR #72, opened 2015-06-26
[SERVICES-423] Remove shadow user creation

ArturKlajnerok:
We don't need anymore functionality of user creation in shadow mode. This introduces two changes :  
[x] Removal of separate user storage for shadow mode
[x] Removal of flag that disables full user creation

/cc @Wikia/services-team


nmonterroso:
LGTM

PR #73, opened 2015-06-27
[SERVICES-473] Defualt cookie path

ArturKlajnerok:
Cookie path need to be set by default to '/' and not depend on the path on which it is created.

/cc @Wikia/services-team


drsnyder:
lgtm

PR #75, opened 2015-06-30
Services 356 save user preferences during registration

sqreek:
becuse of [swagger generator](https://github.com/yvasiyarov/swagger/blob/6a9e6d1cf68a814ebcab98b7405895dd2bc66bd6/parser/operation.go#L148) we cannot use quotes (`"`) in description


nmonterroso:
Cool. For now I think this is fine, but eventually I think we want to improve the Helios Swagger docs to include models.

LGTM

PR #76, opened 2015-06-30
Make the marketingallowed preference explicit

drsnyder:
Previously we were accepting arbitrary preferences encoded in JSON in the `preferences` field in the form submission payload. This PR the marketingallowed preference explicit in the form payload and adds validation for the values accepted by that field.

https://wikia-inc.atlassian.net/browse/SERVICES-480

/cc @Wikia/services-team


common/test.go
```diff
@@ -22,18 +22,14 @@ const (
 	ClientIdKey     = "client_id"
 	ClientSecretKey = "client_secret"
 
-	BirthDateKey = "birthdate"
-	EmailKey     = "email"
-	PassowordKey = "password"
-	UsernameKey  = "username"
```
drsnyder: I'm consolidating these in `common/constants.go`.

domain/user_validator_funs.go
```diff
@@ -10,25 +10,19 @@ import (
 	"github.com/Wikia/helios/utils"
 )
 
-const (
-	UsernameField  = "username"
-	PasswordField  = "password"
-	EmailField     = "email"
-	BirthdateField = "birthdate"
-)
```
drsnyder: Consolidating in `common`.

domain/user_validator_funs.go
```diff
@@ -119,3 +115,21 @@ func ValidateBirthday(user User) (bool, *errorlist.KeyedError) {
...
+
+	for _, preference := range user.Preferences() {
+		if preference.Name == common.MarketingAllowedField {
+			if preference.Value == "0" || preference.Value == "1" {
+				return true, nil
+			} else {
+				return false, ErrInvalidMarketingAllowedPreference
+			}
+		}
+	}
```
drsnyder: This will be a little kludgy until we have more preferences. At that point, we should probably convert them to a map then apply multiple preference checks to that map.

models/user.go
```diff
@@ -18,12 +18,7 @@ const (
 	DefaultOptions         = ""
 	MinHashSize            = 32
 
-	UserIdField     = "user_id"
-	UsernameField   = "username"
-	PasswordField   = "password"
-	EmailField      = "email"
-	BirthdateField  = "birthdate"
-	PreferenceField = "preferences"
```
drsnyder: Consolidated these in `common`.

docs/docs.go
```diff
@@ -604,15 +604,15 @@ var ApiDescriptionsJson = map[string]string{"username":`{
                         },
                         {
                             "paramType": "form",
-                            "name": "preferences",
-                            "description": "Preferences of user, json format, list of preference with name and value property",
-                            "dataType": "string",
-                            "type": "string",
+                            "name": "marketingallowed",
```
ArturKlajnerok: Please add the changes in annotations and generate `docs.go` with `./swagger` script

domain/user_validator_funs.go
```diff
@@ -119,3 +115,21 @@ func ValidateBirthday(user User) (bool, *errorlist.KeyedError) {
 
 	return true, nil
 }
+
+func ValidateMarketingAllowedPreference(user User) (bool, *errorlist.KeyedError) {
+	if user.Preferences() == nil || len(user.Preferences()) == 0 {
```
ArturKlajnerok: This is unneeded if `[]UserPreference` will be `nil` or `empty`, then `for ... range` will be skipped and error will be returned at the and of function.

ArturKlajnerok:
With some small improvements, lgtm :+1:

docs/docs.go
drsnyder: Fixed in 4c1c976.

domain/user_validator_funs.go
drsnyder: Removed in 4c1c976.

PR #77, opened 2015-06-30
Update the swagger annotations in the fb controller

drsnyder:
Change the "preference" annotation to marketingallowed in the fb controller. Both controllers use the same method so the plumbing is the same.

https://wikia-inc.atlassian.net/browse/SERVICES-480

/cc @ArturKlajnerok @nmonterroso


ArturKlajnerok:
Lgtm :+1:

PR #78, opened 2015-07-02
return how the user authenticated

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-439

Add `auth_method` to the Helios `/token` response to indicate how a user was authenticated. This can be used by the client to react, for example by showing a reset password dialog when `auth_method=models.AuthenticatedTempPass`

Related: Wikia/app#7747


models/user.go
```diff
@@ -76,9 +85,16 @@ func (user *User) IsValid() bool {
...
 
 // FIXME: This is not really validation but authentication.
-func (user *User) IsValidPassword(password string) bool {
-	return user.passwordMatchesHash(password, user.ResetPasswordHashed) ||
-		user.passwordMatchesHash(password, user.PasswordHashed)
+func (user *User) AuthenticateUserPassword(password string) int {
+	result := AuthenticateFailed
+
+	if user.passwordMatchesHash(password, user.ResetPasswordHashed) {
+		result = AuthenticatedTempPass
```
ArturKlajnerok: you can return here

models/user.go
```diff
@@ -76,9 +85,16 @@ func (user *User) IsValid() bool {
...
+	} else if user.passwordMatchesHash(password, user.PasswordHashed) {
+		result = AuthenticatedNormalPass
```
ArturKlajnerok: you can return here

ArturKlajnerok:
Other than few small things, lgtm :+1:

PR #80, opened 2015-07-02
Helios accept "on" as an opt-in value for checkbox

tiffanyperumpail:
@nmonterroso https://wikia-inc.atlassian.net/browse/SERVICES-484

This PR adds getCheckBoxValueFromRequest method to support checkbox values of "on" or "empty"


nmonterroso:
+ @Wikia/services-team

ArturKlajnerok:
One of the tests is failing : `FAIL: TestE2eUsersValidUserWithMarketingAllowedOn`

bkoval:
What about a situation when there's no "marketingallowed" field sent in POST data? It's the default behavior in the browser for unchecked checkboxes to strip it down from the data that is sent to the remote resource and currently this situation is not being handled by Helios.

tiffanyperumpail:
@nmonterroso  Changes are finalized. This should cover marketingallowed set to "on" or no marketingallowed field in a POST. Let me know if anything needs to be fixed.

ArturKlajnerok:
Lgtm :+1:

ArturKlajnerok:
@bkoval With the changes introduced in this PR lack of `marketingallowed` parameter in request is setting it to `false/0` by default.

PR #81, opened 2015-07-10
[SERVICES-481] Setting language preference based on language code

ArturKlajnerok:
With latest changes to sending email that use language based on user preferences we need to set it on user creation.

/cc @Wikia/services-team


nmonterroso:
LGTM

PR #82, opened 2015-07-13
[SERVICES-481] Confirmation email language cleanup

ArturKlajnerok:
We don't need to send language as parameter of account confirmation email request. Language is now set based on user preference.

/cc @Wikia/services-team


nmonterroso:
LGTM - the corresponding changes are already live on the MW side?

ArturKlajnerok:
@nmonterroso Changes for supporting email language based only on preferences were introduced while ago by the social team.  There is additional cleanup on MW app side : https://github.com/Wikia/app/pull/7831 but they don't depend on each other.

PR #83, opened 2015-07-15
[SERVICES-421] Proper length checking based on runes

ArturKlajnerok:
Length of username and password should be counted based on number of runes.

/cc @Wikia/services-team


nmonterroso:
LGTM

PR #84, opened 2015-07-15
language input validation

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-526

Add some simple validation to the language of a user.


common/constants.go
```diff
@@ -5,6 +5,7 @@ const (
 	MinAge              = 13
 	MinLen              = 1
 	MaxLen              = 50
+	LanguageMaxLen      = 8
```
ArturKlajnerok: I think that for most language code standards 5 would be a good max, but MW app support one that has 12 characters : https://github.com/Wikia/app/blob/808a769df6cf8524aa6defcab4f971367e3e3fd8/languages/Names.php#L384 :crying_cat_face:

ArturKlajnerok: Considering there's only small change with max length, LGTM :+1:

nmonterroso: Good find, updated in 97a4f97

ArturKlajnerok:
Lgtm :+1:

PR #85, opened 2015-07-22
Swagger Updates

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-468

Change from Swagger annotations to Swagger doc written in the [Swagger Editor](http://editor.swagger.io/#/) since the annotations don't support everything we need (authorizations) and we noticed some limitations on the models generated by the tool. Also updating to the 2.0 spec.

~~I had to change the `swagger_controller` tests to be integration tests, since it serves a static file using a relative location from the working dir (`github.com/Wikia/helios`), but since coverage is run during jenkins tests, the tests are actually copied to `github.com/Wikia/helios/controllers/_test/.....`, the newly generated test file is not able to read the file contents as I expected.~~


ArturKlajnerok:
Lgtm :+1:

PR #86, opened 2015-07-27
SERVICES-531 Add "POST" handler to FB Connect

bkoval:
https://wikia-inc.atlassian.net/browse/SERVICES-531

@ArturKlajnerok @nmonterroso Since we're not sure if PUT requests and CORS work nicely in the browsers, how about we add a POST handler to FB Connect resource to ensure the requests are not blocked?


kvas-damian:
LGTM

drsnyder:
> Since we're not sure if PUT requests and CORS work nicely in the browsers, how about we add a POST handler to FB Connect resource to ensure the requests are not blocked?

We decided upon `PUT` here because the URI `/users/<user_id>/facebook_app_id/<app_id>` can and probably will be used as a stable resource for the connection whose representation provides metadata about the connection. I would rather not compromise the HTTP semantics until we know for sure that we must in order to get around CORS or some other issue.

drsnyder:
@ArturKlajnerok is currently looking into whether or not the api-gateway is the issue. Let's see what he finds out.

ArturKlajnerok:
Closing as we solve the CORS problem on the api-gateway level.

PR #88, opened 2015-07-29
[SERVICES-377] Returning access token after account creation.

ArturKlajnerok:
Access token in response body and in cookie should be returned after successful account creation.

- [x] Create common way of setting access token and cookie to the response.
- [x] Return access token/set cookie from `/user` resource
- [x] Return access token/set cookie from `/facebook/users` resource
- [x] Adjust access token/set cookie in `/facebook/token` resource
- [x] Update swagger docs

/cc @Wikia/services-team


resources/swagger/swagger.json
```diff
@@ -522,6 +524,58 @@
                 }
             }
         },
+        "TokenSuccessResponse": {
```
nmonterroso: I wonder if this should just extend `TokenResponse` with non-`required` properties `success` and `success_description`, that way we don't have to keep both models in sync. For example, `TokenResponse` now doesn't have enums 5 and 6 for `auth_method`s

ArturKlajnerok: I've extended `TokenResponse ` instead of creating new response in : 70fe9f8

nmonterroso:
LGTM

controllers/helpers.go
```diff
@@ -85,6 +96,54 @@ func getCheckBoxValueFromRequest(req *http.Request) string {
...
+	if exists {
+		cookie := createAccessTokenCookie(accessToken.(string), domain)
+		http.SetCookie(responseWriter, &cookie)
+	}
+}
+
+func createAccessTokenCookie(accessToken string, domain string) http.Cookie {
+	cookieExpireDate := time.Now().AddDate(10, 0, 0) // 10 years
+	return http.Cookie{Name: AccessTokenKey, Value: accessToken, Path: DefaultCookiePath, Domain: domain, Expires: cookieExpireDate, HttpOnly: true}
+}
```
drsnyder: Minor: I think we should put these somewhere other than the helper. This might belong in a package all it's own.

controllers/helpers.go
```diff
@@ -85,6 +96,54 @@ func getCheckBoxValueFromRequest(req *http.Request) string {
...
+	handleMultipleAccessTokens(controller, userIdStr, accessRequest)
+
+	controller.GetServer().FinishAccessRequest(resp, req, accessRequest)
+	if accessRequest.Authorized == true {
+		resp.Output[models.UserIdField] = userIdStr
+		resp.Output[models.AuthMethodField] = authMethod
+	}
+}
+
+func handleMultipleAccessTokens(controller AccessController, userId string, accessRequest *osin.AccessRequest) {
```
drsnyder: The meaning of this function is not clear to me. With the name `Multiple` I was expecting a loop.

controllers/helpers.go
```diff
@@ -85,6 +96,54 @@ func getCheckBoxValueFromRequest(req *http.Request) string {
...
+	if !controller.GetConfig().Server.AllowMultipleAccessTokens {
+		accessData, err := controller.GetRedisStorage().GetAccessForUserId(userId)
+		if err == nil && accessData != nil {
+			accessRequest.ForceAccessData = accessData // Reuse previous token if it exists
```
drsnyder: I don't think we should modify the state of the `accessRequest object` directly here.

controllers/helpers.go
```diff
@@ -85,6 +96,54 @@ func getCheckBoxValueFromRequest(req *http.Request) string {
...
+			Id:          controller.GetConfig().Client.ClientId,
+			Secret:      controller.GetConfig().Client.ClientSecret,
+			RedirectUri: controller.GetConfig().Client.RedirectUri,
+			UserData:    nil,
+		},
+	}
+
+	handleMultipleAccessTokens(controller, userIdStr, accessRequest)
+
+	controller.GetServer().FinishAccessRequest(resp, req, accessRequest)
```
drsnyder: This violates the "Only talk to your immediate friends" portion of the [law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter). That is, it dilutes the purpose of the `AccessController` interface because there is a concrete implementation coming from `GetServer` and `addAccessTokenToResponse` doesn't have direct access to `FinishAccessRequest`. The same issue applies to the `controller.GetConfig().Client.*` calls above and `controller.GetRedisStorage().GetAccessForUserId()` below.

I think we should spend a little bit of time designing an "auth" interface that should satisfy the different call sites that need to generate a token. When we have that then we can inject whatever the auth object needs. This should also allow us to abstract `osin` behind this auth interface so we move further along towards decoupling our responses from it.

PR #90, opened 2015-08-03
Change User Password 

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-549

Provides the authenticated resource `PUT /users/{userId}/password` to change a user's password.


drsnyder:
lgtm

PR #91, opened 2015-08-05
[SERVICES-557] Confirm valid emails based on facebook accounts

ArturKlajnerok:
We should confirm valid user emails that comes from facebook user creation and not send them confirmation email as part of post creation process.

 * [x]  Confirm emails with facebook 
 * [x]  Don't send confirmation email if email is verified
 * [x]  Set post creation commands based on domain user status
 * [x] Adjust tests


drsnyder:
Description?

drsnyder:
What ticket is this associated with?

controllers/facebook_controller.go
```diff
@@ -70,6 +72,9 @@ func NewFacebookController(
 	userValidator := domain.NewUserValidator(validationFunctions...)
 	controller.userService = service.NewUserService(userStorage, userValidator, wikiaClient, userPreferenceGateway)
 
+	controller.commandsConfirmedEmail = []domain.ApplyToUserFunc{controller.userService.UpdateAntiSpoofCommand}
+	controller.commandsAll = []domain.ApplyToUserFunc{controller.userService.UpdateAntiSpoofCommand}
```
drsnyder: They are the same-- are they both necessary?

controllers/facebook_controller.go
```diff
@@ -156,3 +162,12 @@ func (controller *FacebookController) usersHandler(w http.ResponseWriter, r *htt
...
 }
+
+func (controller *FacebookController) getProperUserPostCreationCommands(emailConfirmed bool) []domain.ApplyToUserFunc {
+	if emailConfirmed {
+		return controller.commandsConfirmedEmail
+	} else {
+		return controller.commandsAll
+	}
+
+}
```
drsnyder: Could we encapsulate this logic in `SendConfirmationEmailCommand`? If `user.EmailConfirmed` that command could be a noop.

ArturKlajnerok:
@drsnyder:  Ticket number in the PR title : SERVICES-557

controllers/facebook_controller.go
ArturKlajnerok: I didn't assigned properly `controller.commandsAll` fixed in : 9ef0fe7

drsnyder:
Derp. Sorry about that.

controllers/facebook_controller.go
ArturKlajnerok: I've did that in bc89e44, but I'm not sure if this is makes it more readable :confused:

domain/user.go
```diff
@@ -45,33 +48,42 @@ func NewUser(name, password, email string, birthday interface{}) *UserValue {
...
 }
 
-func (user UserValue) Birthday() *time.Time {
+func (user *UserValue) EmailConfirmed() bool {
+	return user.emailConfirmed
+}
+
+func (user *UserValue) SetEmailConfirmed(confirmed bool) {
+	user.emailConfirmed = confirmed
+}
```
drsnyder: Are we sure we want this to be mutable?

controllers/users_controller.go
```diff
@@ -80,6 +81,9 @@ func NewUsersController(
 	controller.facebookService = service.NewFacebookService(
 		controller.userStorage, gateway.NewFacebookGateway(facebookConfig), controller.userService)
 
+	controller.userPostCreationCommands = []domain.ApplyToUserFunc{
+		controller.userService.UpdateAntiSpoofCommand, controller.userService.SendConfirmationEmailCommand}
+
```
drsnyder: This looks duplicated with the facebook controller.

service/user.go
```diff
@@ -129,6 +120,10 @@ func (service *UserServiceValue) UpdateAntiSpoofCommand(user domain.User) (bool,
 }
 
 func (service *UserServiceValue) SendConfirmationEmailCommand(user domain.User) (bool, *errorlist.KeyedError) {
+	if user.EmailConfirmed() {
+		return true, nil
+	}
+
```
drsnyder: I think the existing post creation command logic could remain the same since this effectively short circuits sending the confirmation email.

drsnyder: That is, I don't think we need the changes in the controllers.

gateway/facebook.go
```diff
@@ -10,9 +10,16 @@ import (
 )
 
 const (
+	FacebookEmailResource = "/me?scope=email"
```
nmonterroso: does this require an additional permission when a user links their facebook account?

ArturKlajnerok: Yes, by default user email is shared but user can unshare it during the authorization process. Proper scope `scope=email` is needed to get that data if they are accessible.

service/user.go
ArturKlajnerok: I've moved commands back to the user service in : 1393556

PR #93, opened 2015-08-07
Provide a client for sending the reset password email through MediaWiki

drsnyder:
I'm creating this to get the conversation started about the structure of the code for the reset password flow. This PR creates a failing test for `ResetPassword(name, auth)` that uses the client to media wiki.

With the straw man in place I'm going to:
 * [x] Fill in `SendPasswordResetEmail`
 * [x] Fill in `GenerateTemporaryPassword`
 * [x] Fix the remaining tests that are failing
 * [x] Determine what needs to be done to make a MW request "internal" (the corresponding MW change is [here](https://github.com/Wikia/app/pull/8059))

https://wikia-inc.atlassian.net/browse/SERVICES-587

/cc @Wikia/services-team


errorlist/vars.go
```diff
@@ -2,5 +2,6 @@ package errorlist
 
 var (
 	GenericDataAccessError = NewKeyedError("server", "server_error")
+	GenericSystemError     = NewKeyedError("server", "internal_server_error")
```
ArturKlajnerok: Should we introduce `internal_server_error` as for now all internal problems with the service returned `server_error`?

ArturKlajnerok:
Other than one small question, lgtm :+1:

errorlist/vars.go
drsnyder: Ah, if `server_error` is how we are handling that, I'll stick with the convention.

drsnyder: Fixed in e6b9ae7.

PR #94, opened 2015-08-12
Properly set the user's ID in Save

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-591

When we changed the `Save` function in `UserStorage` to be an `INSERT .... ON DUPLICATE KEY UPDATE`, we lost the part of the `Insert` db function that set the user's ID in the model. Luckily the `Exec` call can get us the ID, either from an insert or update.


ArturKlajnerok:
Lgtm :+1:

PR #95, opened 2015-08-13
[SERVICES-615] Added underscore to illegal username characters

ArturKlajnerok:
/cc @Wikia/services-team @owend


nmonterroso:
Is it that usernames are blocked? I thought I could ask for username `hi_there`, but MW would transform it to `hi there`?

ArturKlajnerok:
@nmonterroso I've double check that.

sqreek:
LGTM

ArturKlajnerok:
Ok, it seems that it's safe to block that character instead of transforming it to space.

PR #96, opened 2015-08-14
[SERVICES-583] Reset password resource

ArturKlajnerok:
This is resource responsible for starting reset password flow for the user. It's providing temporary password sent via email that can be later used with other resource to change the main password.

 * [x] Create a controller for resource that uses user service to start the reset password flow.
 * [x] Throttle getting temporary password for user to one per 24h
 * [x] Save temporary password to storage

/cc @Wikia/services-team


drsnyder:
Please provide a description.

controllers/users_controller.go
```diff
@@ -94,6 +96,9 @@ func NewUsersController(
 		"/users/{user_id:[0-9]+}/password",
 		timedCorsHandler(controller, "passwordChange", MuxVarsHandler(controller.changeUserPassword).ToHandlerFunc())).Methods("PUT")
 
+	rtr.HandleFunc("/users/reset_password",
```
drsnyder: I wander if there isn't a better URI for this. The weird part is that we normally accept a user id on the URI and in this case we only have a username.

What about `POST /users/{username}/password`?

controllers/users_controller.go
```diff
@@ -109,6 +114,18 @@ func (controller *UsersController) GetMetricsClient() *client.Client {
...
+	return controller.config
+}
+
+func (controller *UsersController) GetServer() *osin.Server {
+	return controller.server
+}
+
+func (controller *UsersController) GetRedisStorage() *storage.RedisStorage {
+	return controller.redisStorage
+}
```
drsnyder: Where are these used? Providing these seems like a potential code smell or tight coupling to another part of the code base.

controllers/users_controller.go
ArturKlajnerok: In this case a username value is a requirement, and resource should be usable only with that parameter. I think we can change it to `POST /users/{username}/password` if this makes it more consistent.

controllers/users_controller.go
ArturKlajnerok: These methods were introduced when we added the support for creating tokens from different controller: https://github.com/Wikia/helios/blob/master/controllers/helpers.go#L99 We can clean that up with introduction of a proper auth service that will encapsulate access token creation and requests to osin.

models/user.go
```diff
@@ -90,6 +90,37 @@ func (user *User) IsValid() bool {
...
+	newUser.PasswordHashed = hash
+	return newUser
+}
+
+func (user *User) SetResetPasswordHashed(hash string) *User {
+	newUser := user
+	newUser.ResetPasswordHashed = hash
+	newUser.ResetPasswordTime = utils.GetCurrentFormatedTime()
+	return newUser
+}
```
drsnyder: Do these return copies?

service/user.go
```diff
@@ -98,20 +100,40 @@ func (service *UserServiceValue) ChangePassword(id int64, password string) (bool
...
+		errors.Add(UserIsBlockedError)
+		return false, errors
+	}
+
+	if user.IsTemporaryPasswordThrottled() {
+		errors.Add(ThrottledPasswordReset)
+		return false, errors
+	}
+
 	tempPassword, passwordCreationErrors := service.authService.GenerateTemporaryPassword(common.TemporaryPasswordSize)
```
drsnyder: In hindsight, maybe `GenerateTemporaryPassword` should have been in `utils`.

drsnyder:
Minor comments, otw LGTM.

wikia/wikia.go
```diff
@@ -127,11 +134,11 @@ func (client *Client) SendPasswordResetEmail(name, newPassword string) (bool, er
 
 	response := &Response{}
 	_, err := client.callWithToken(endpoint, data, response)
-	if err != nil {
+	if err != nil || response.Result != ResultSuccess {
```
sqreek: Can be this replaced by
`return !(err!=nil || response.Result != ResultSuccess), err`
or it's out of convention?

nmonterroso:
Can you run `./generateResources.sh`? otherwise LGTM

controllers/users_controller.go
ArturKlajnerok: @drsnyder Changed in : 861f2e8

models/user.go
ArturKlajnerok: Good catch, in this case to make a proper copy pointer needed to be dereferenced, fixed in : 39a784a

service/user.go
ArturKlajnerok: For mocking purposes it makes sense to keep `GenerateTemporaryPassword` in `AuthService`.

wikia/wikia.go
ArturKlajnerok: This should also work, but it's out of current convention.

PR #97, opened 2015-08-14
Add IsUsernameBlocked client request to Helios

drsnyder:
The corresponding MediaWiki change is [here](https://github.com/Wikia/app/pull/8098). This adds the client functionality via `IsUsernameBlocked` in the `wikia` package.

https://wikia-inc.atlassian.net/browse/SERVICES-595

/cc @Wikia/services-team


wikia/wikia.go
```diff
@@ -118,10 +126,39 @@ func (client *Client) SendPasswordResetEmail(name, newPassword string) (bool, er
...
+	data.Add(usernameKey, name)
+
+	payload := &BlockedResponse{}
+	response, err := client.callWithToken(endpoint, data, response)
+	if err != nil {
+		return false, err
+	}
+
+	if response.StatusCode == http.StatusOK {
+		return payload.Blocked, nil
```
nmonterroso: does this have to be json_decoded into? right now it creates a `BlockedResponse` but never writes to it, so it should always be whatever the default for a `bool` is (false?)

nmonterroso:
are tests part of a different ticket?

ArturKlajnerok:
Lgtm :+1:

wikia/wikia.go
drsnyder: Oh, interesting. Good catch. This is a typo. I guess that's the problem with passing a writable interface to the function.

drsnyder:
> are tests part of a different ticket?

Good call. Currently, none of the code in this package is tested.

Let's change that.

wikia/wikia.go
drsnyder: Fix this specific issue in 9fed23c. I created https://wikia-inc.atlassian.net/browse/SERVICES-622 for improving the testability of this package.

drsnyder:
In standup we agreed to merge this. It will be used with what @ArturKlajnerok is implementing to complete the password reset flow.

PR #98, opened 2015-08-21
Services-462 invalidate user access token

harnash:
Main task was to add ability to invalidate user access tokens through Helios. Here is the story behind this commit.

1. In controllers there were places where simple http.HandleFunc() was used (oauth_controller) or gorilla/mux.Router (users_controller). Since for tokens we would need more complex route handling it was logical to use gorilla's router there. Problem appeared that it seems that declaring multiple routers is not so simple and it is not very intuitive and [documentation](http://www.gorillatoolkit.org/pkg/mux) wasn't very helpful. I found a [post](http://trevorgattis.blogspot.com/2014/07/looking-at-gorillas-mux-i-wanted-to.html) which kinda helped but I ran into different issues. In the end we have routers in (oauth_controller and users_controller) and I had to add controllers.GetRouter() to allow registering all routes to the same (main) router.

2. Endpoint for invalidating resources looks like this:
`[DELETE]: /token/{user_access_token}`
it doesn't do any authentication and always returns 200 (or 5xx in case of exception). It will invalidate all authentication entries (if token is the access token). In case where user pass refresh_token it will try to find access token assigned to this refresh token and will delete both tokens. There is no way at this moment to delete all refresh tokens having only access token.

3. Since debugging helios running locally is quite painful I have added option to send all logs to STDOUT which should be used for development purposes only.

TODO: 
* add routers to other resources/controllers
* ~~write tests for new endpoint~~
* ~~update swagger docs~~

ping: @Wikia/services-team


harnash:
@ArturKlajnerok: please take a look at my tests failing... I can't find any reason for the error compiler is reporting
```
./test ./...
# _/Users/harnas/_Projects_/helios/controllers
controllers/helpers.go:192: redisStorage.RemoveAccessTokenForUserId undefined (type *storage.RedisStorage has no field or method RemoveAccessTokenForUserId)
?   	_/Users/harnas/_Projects_/helios	[no test files]
?   	_/Users/harnas/_Projects_/helios/common	[no test files]
?   	_/Users/harnas/_Projects_/helios/config	[no test files]
FAIL	_/Users/harnas/_Projects_/helios/controllers [build failed]
ok  	_/Users/harnas/_Projects_/helios/domain	0.015s
ok  	_/Users/harnas/_Projects_/helios/errorlist	0.012s
--- FAIL: TestUpdateUserPreference (0.11s)
        Location:       user_preference_e2e_test.go:30
	Error:		Should be true
	Messages:	failed to update

        Location:       user_preference_e2e_test.go:31
	Error:		Expected nil, but got: &errorlist.KeyedError{Key:"server", Message:"user_preference_service_error"}
	Messages:	An Error occur

--- FAIL: TestUserPreferenceServiceE2eTestSuite (0.11s)
FAIL
FAIL	_/Users/harnas/_Projects_/helios/gateway	0.125s
?   	_/Users/harnas/_Projects_/helios/helpers	[no test files]
ok  	_/Users/harnas/_Projects_/helios/models	0.013s
ok  	_/Users/harnas/_Projects_/helios/problem	0.009s
?   	_/Users/harnas/_Projects_/helios/resources	[no test files]
?   	_/Users/harnas/_Projects_/helios/server	[no test files]
ok  	_/Users/harnas/_Projects_/helios/service	0.015s
?   	_/Users/harnas/_Projects_/helios/storage	[no test files]
ok  	_/Users/harnas/_Projects_/helios/utils	0.017s
?   	_/Users/harnas/_Projects_/helios/wikia	[no test files]
godep: go exit status 2
```

harnash:
@ArturKlajnerok: OK. Upgrading to latest go (1.5) solved this problem.

nmonterroso:
does that mean we have to upgrade `go` to 1.5 before this change goes out?

harnash:
@nmonterroso I think we are running 1.4.1 in dev and perhaps in prod and it is running fine (tests). We should schedule upgrading to 1.5 since it does improve garbage collector etc. but we don't have to do it now. I had some very old version 1.3 so I bet it was problem with my dev environment.

controllers/oauth_controller.go
```diff
@@ -44,10 +44,15 @@ func NewOAuthController(
...
-	http.HandleFunc("/info",
-		timedCorsHandler(controller, "infoHandler", controller.infoHandler))
-	http.HandleFunc("/token",
-		timedCorsHandler(controller, "tokenHandler", controller.tokenHandler))
+	GetRouter().HandleFunc("/info",
+		timedCorsHandler(controller, "infoHandler", controller.infoHandler)).Methods("POST")
+
+	token_router := GetRouter().PathPrefix("/token").Subrouter()
+
+	token_router.HandleFunc("/",
```
nmonterroso: does this mean that this responds only to `/token/` but not `/token`?

nmonterroso:
Please remember to update the swagger spec (instructions defined in the `README`)

drsnyder:
> `[DELETE]: /token/{user_access_token}` it doesn't do any authentication and always returns 200 (or 5xx in case of exception). 

Shouldn't this resource require authentication?

harnash:
@drsnyder Why should it? You are sending token which is valid (or not) and this is the source of authentication. We can't really check user identity unless we require him to send all credentials again which I think is pointless in this case.

harnash:
@Wikia/services-team 
I think this is ready for merge and routers can be added in other tickets which are currently blocked by this PR (SERVICE-623).

drsnyder:
> Why should it? You are sending token which is valid (or not) and this is the source of authentication. We can't really check user identity unless we require him to send all credentials again which I think is pointless in this case.

@harnash if we do this (don't authenticate) we are opening ourselves up to brute force attacks on our tokens. An attacker could fairly easily enumerate the universe of tokens and proceed to delete them one by one. It's not a huge risk, but it's fairly easy to prevent.

harnash:
I have added user_id verification however I have no time to modify tests (`TestE2eOauthInvalidateExistingToken`, `TestE2eOauthInvalidateNonExistingToken`) and add new one to test if this code works properly. This is kinda problematic since tests do not send headers anywhere atm.

harnash:
After having a chance to rethink this authorization I think that requiring additional header to invalidate token doesn't improve security that much. Enumerating tokens is quite resource consuming and requires constant access to resource (which can and should be easily restricted with rate limiting/ip blocking etc.). It undermines access token as the only source of user authentication for the app. If we were going to use OAuth2 scheme we should send client_id and client_secret but that's not the case? There for I'm for removing this restriction.

drsnyder:
> After having a chance to rethink this authorization I think that requiring additional header to invalidate token doesn't improve security that much. Enumerating tokens is quite resource consuming and requires constant access to resource (which can and should be easily restricted with rate limiting/ip blocking etc.).

The tokens are just fixed length strings of letters and numbers. Enumerating them is just a brute force task.

Requiring a valid session means that someone needs a valid, authorized token in order to delete the resource. That means that someone would have to go so far as to login and get their token before they started the brute force attack. I will make the request (more) traceable and the attack more difficult.

harnash:
Bruteforce attack are easily mitigated by rate limiting and black/whitelisting or am I missing something? Access token IS the source of authorization so I can theoretically  guess some one's token and proceed as logged in user anyway (kind of session hijacking). What keeps me from sending user_id in header anyway (still guessed but from point of view Helios unverifiable)? Isn't header value (user_id) extracted from access token anyway?
I'm just trying to simplify things and understand how this should work :-)

drsnyder:
> Bruteforce attack are easily mitigated by rate limiting and black/whitelisting or am I missing something?

Yes. We should do that too.

> Access token IS the source of authorization so I can theoretically guess some one's token and proceed as logged in user anyway (kind of session hijacking). What keeps me from sending user_id in header anyway (still guessed but from point of view Helios unverifiable)?

If we require a valid access token as part of the request then you either have to login first and get your access token, hijack another session or guess a valid token before you can start the brute force attack. Requiring any one of those will make it a lot harder to initiate a successful attack.

Simply providing the `X-Wikia-UserId` in the headers is not sufficient. The API gateway [strips it](https://github.com/Wikia/api-gateway/blob/master/src/nginx.lua#L56-L69) and only supplies it to the backing service if there is a valid token either in the `access_token` cookie or in `X-Wikia-AccessToken` header.

> Isn't header value (user_id) extracted from access token anyway?

Yes, but as a precaution, we should make sure that the user requesting to delete the token is in fact the owner of the token and by requiring the request to be authenticated at least raise the bar to make sure the requestor is following the "protocol".

harnash:
Here is some additional PR somehow related with this one: https://github.com/Wikia/go-commons/pull/13

controllers/helpers.go
```diff
@@ -134,13 +140,82 @@ func handleMultipleAccessTokens(controller AccessController, userId string, acce
...
+		logger.GetLogger().Debug("Could not find access data for token - trying to find refresh token")
+		accessData, err = redisStorage.LoadRefresh(token)
+		if err != nil || accessData == nil {
+			logger.GetLogger().Error("Could not find access data for token")
+			return nil
+		}
+
+		refreshToken = token
+		token = accessData.AccessToken
+	}
```
nmonterroso: should there be an 
```go
else {
  refreshToken = accessData.RefreshToken
}
```

here? that way we can delete it below, so we're consistently deleting both the refresh token and access token when a user logs out, especially since I think the common case here will be to provide access tokens instead of refresh tokens, it will help reduce how much memory Helios consumes in Redis.

harnash: Good catch! I missed that. Fixed in 3a1c509.

resources/swagger/swagger.json
```diff
@@ -96,7 +96,7 @@
                     "token"
                 ],
                 "summary": "Oauth2 Resource Owner Password Credentials",
-                "operationId": "Token",
+                "operationId": "TokenGenerate",
```
nmonterroso: I wonder if the operation id here should be `login`, so it's a little bit more obvious what it's doing when looking at the swagger doc

resources/swagger/swagger.json
```diff
@@ -163,6 +163,35 @@
...
+                            "$ref": "#/definitions/SimpleError"
+                        }
+                    }
+                },
+                "description": "invalidates user access token",
+                "tags": [
+                    "token"
+                ],
+                "summary": "Oauth2 Access Token invalidation",
+                "operationId": "TokenInvalidate",
```
nmonterroso: Should this be `logout`?

nmonterroso:
minor comments otw lgtm

resources/swagger/swagger.json
harnash: Changed in fdafd89.

controllers/helpers.go
```diff
@@ -134,13 +140,84 @@ func handleMultipleAccessTokens(controller AccessController, userId string, acce
...
+	if err != nil {
+		logger.GetLogger().Error("Error gettin access data for token")
+		return err
+	}
+
+	refreshToken := ""
+
+	if accessData == nil {
+		// Check if we are dealing with refresh token
+		logger.GetLogger().Debug("Could not find access data for token - trying to find refresh token")
```
drsnyder: Minor: if we are going to log this event we'll probably want the token in the message.

controllers/helpers.go
```diff
@@ -134,13 +140,84 @@ func handleMultipleAccessTokens(controller AccessController, userId string, acce
...
+	if err != nil {
+		logger.GetLogger().Warn(fmt.Sprintf("Error deleting access token (access token): %v", err))
+		return err
+	}
+
+	err = redisStorage.RemoveAccess(token)
+	if err != nil {
+		logger.GetLogger().Warn(fmt.Sprintf("Error deleting access token (access): %v", err))
+		return err
+	}
```
drsnyder: I don't think this function belongs in the controller. It should probably be in the `auth` service.

drsnyder:
Minor comments, OTW, LGTM

controllers/helpers.go
```diff
@@ -134,13 +140,84 @@ func handleMultipleAccessTokens(controller AccessController, userId string, acce
...
+		err = redisStorage.RemoveRefresh(refreshToken)
+		if err != nil {
+			logger.GetLogger().Warn(fmt.Sprintf("Error deleting access token (refresh): %v", err))
+			return err
+		}
+	}
+
+	logger.GetLogger().Debug(fmt.Sprintf("Deleting token for user: %v", userId))
+
+	err = redisStorage.RemoveAuthorize(userId)
```
ArturKlajnerok: We don't need to remove `Authorize` as we don't support this flow and it's never used. Methods for `Authorize` are implemented only to meet the interface for storage.

controllers/helpers.go
```diff
@@ -134,13 +140,84 @@ func handleMultipleAccessTokens(controller AccessController, userId string, acce
...
+	if err != nil {
+		logger.GetLogger().Warn(fmt.Sprintf("Error deleting access token (authorize): %v", err))
+		return err
+	}
+
+	err = redisStorage.RemoveAccessTokenForUserId(accessData)
+	if err != nil {
+		logger.GetLogger().Warn(fmt.Sprintf("Error deleting access token (access token): %v", err))
```
ArturKlajnerok: This is removing a `user_id->access_token` mapping so I would change the log message in this case.

controllers/helpers.go
```diff
@@ -134,13 +140,84 @@ func handleMultipleAccessTokens(controller AccessController, userId string, acce
...
+		return err
+	}
+
+	err = redisStorage.RemoveAccess(token)
+	if err != nil {
+		logger.GetLogger().Warn(fmt.Sprintf("Error deleting access token (access): %v", err))
```
ArturKlajnerok: Should we also use a log level `Error`here ?

controllers/helpers.go
harnash: Fixed in b8b19a8.

controllers/helpers.go
harnash: Fixed in b8b19a8.

controllers/helpers.go
harnash: Fixed in b8b19a8.

controllers/helpers.go
harnash: Fixed in b8b19a8.

controllers/helpers.go
harnash: Fixed in e8cb2f7.

drsnyder:
lgtm

PR #99, opened 2015-08-24
Updated test user password after mass-password-change

harnash:
After recent mass-change of dev user passwords `Heliostest` user stopped being authenticated in tests. Updated user password and changed it test configuration.

**This is quite urgent since our tests are failing on master.**

@Wikia/services-team


harnash:
I have no idea why some tests are failing on Jenkins... Most of the tests failing there are passing locally on my dev machine :(

ArturKlajnerok:
@harnash  Tests on Jenkins run with SJC config, so we need to be sure that data for test user are valid in  dev db. It's also possible that db config on Jenkins is outdated as there were some changes lately with dev db setup.

harnash:
@ArturKlajnerok I see. So we should update user credentials on dev-sjc and it would work fine then? There are also some issues with consul in tests which I'm investigating now.

drsnyder:
We really need test fixtures.

harnash:
OK. I have updated passwords for POZ and SJC for user `Heliostest`. Since now tests will fail on master I am merging this change.

harnash:
@ArturKlajnerok You were right. `config.ini` in Jenkis Helios test was outdated - fixed that (set to `dev-db-s1`).

PR #100, opened 2015-08-24
prevent account link overwriting

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-620

change facebook account linking to fail when a facebook account link already exists, either for the facebook user or for the wikia user.


harnash:
LGTM although I wished we could do better SQL/query/error handling

nmonterroso:
I agree, I wish that the MySQL package would return a typed "this is a duplicate key" error instead of us having to parse the error string.

PR #102, opened 2015-08-28
fixing sql

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-623

fixing SQL error that was causing an issue when trying to unlink FB accounts from Wikia accounts.


ArturKlajnerok:
Lgtm :+1:

PR #103, opened 2015-08-28
[SERVICES-672] Switched to new temporary password email api call

ArturKlajnerok:
Switched sending email with temporary password to api call that returns different http status codes and sends email in proper format.

/cc @Wikia/services-team


nmonterroso:
assuming the changes are present on the MW side LGTM

PR #104, opened 2015-08-28
Password change cleanup

ArturKlajnerok:
After password is changed temporary password should be removed also the time of temporary password generation should be cleared. In addition I changed few names to make them a bit simpler.

/cc @Wikia/services-team


nmonterroso:
LGTM once tests pass

PR #105, opened 2015-09-01
[SERVICES-682] Fix BirthDateFormat and tests

sqreek:
https://wikia-inc.atlassian.net/browse/SERVICES-682


domain/user.go
```diff
@@ -98,6 +98,10 @@ func BirthDateStringToTime(birthDate string) *time.Time {
 		return nil
 	}
 
+	if date.Format(BirthDateFormat) != birthDate{
```
kvas-damian: missing space ;)

domain/user_validator_test.go
```diff
@@ -17,7 +17,9 @@ const (
 )
 
 var (
-	InvalidBirthdate = fmt.Sprintf("%04d-%02d-%02d", CurrentTime.Year(), CurrentTime.Month(), CurrentTime.Day())
+	TodayBirthdate = fmt.Sprintf("%04d-%02d-%02d", CurrentTime.Year(), CurrentTime.Month(), CurrentTime.Day())
```
pchojnacki: Why this example date needs to be 'today  ?

sqreek: Because we need to check user's age.
And we don't want to fail tests after 13 years.
Long live Helios.

domain/user.go
```diff
@@ -98,6 +98,10 @@ func BirthDateStringToTime(birthDate string) *time.Time {
 		return nil
 	}
 
+	if date.Format(BirthDateFormat) != birthDate {
```
nmonterroso: can we add a note on why this would be possible? It took me a few minutes and looking at the ticket a few times to realize that this would happen because of a date that isn't possible and how that would make this check fail the conversion.

nmonterroso:
minor comment otw LGTM

domain/user_validator_test.go
pchojnacki: ok but please do use CurrentTime.Format("2006- ...

ArturKlajnerok: @pchojnacki Corrected in : 0e6332d

domain/user.go
ArturKlajnerok: @nmonterroso Added comment here : 78dc829

ArturKlajnerok:
Lgtm :+1:

nmonterroso:
LGTM

PR #106, opened 2015-09-01
Facebook integration testing

nmonterroso:
@Wikia/services-team 

Adding `FacebookTestHelper` which allows us to create and remove facebook test users so we can do some integration testing on the facebook resources.

While doing this I realized that `common/test.go` has become a dumping ground for anything test-related that we want shared between tests. I started to separate that out to `helpers/testhelpers` as part of this. If we like that direction we should do the same for everything in `common/test.go`


helpers/facebook.go
```diff
@@ -0,0 +1,9 @@
+package helpers
+import "github.com/huandu/facebook"
+
+func CreateFacebookApp(appId, appSecret string) *facebook.App {
+	app := facebook.New(appId, appSecret)
+	app.EnableAppsecretProof = true
+
+	return app
+}
```
drsnyder: Should this be in the `gateway` package?

nmonterroso: Hm, maybe? I put this alongside [`helpers/osin.go`](https://github.com/Wikia/helios/blob/master/helpers/osin.go) which basically does the same thing for the osin server

helpers/testhelpers/fbtesting/test_users.go
```diff
@@ -0,0 +1,77 @@
...
+
+type FacebookTestHelper struct {
+	app *facebook.App
+	testing *testing.T
+	session *facebook.Session
+}
+
+func NewFacebookTestHelper(t *testing.T) *FacebookTestHelper {
+	conf := config.LoadConfig("../config/config.ini")
+	app := helpers.CreateFacebookApp(conf.Facebook.AppId, conf.Facebook.AppSecret)
```
drsnyder: Can this be mocked?

drsnyder:
Minor comments. I think my only concern is mocking the facebook integration. I think stubbing that part out is sufficient and will allow us to focus on the helios integration with our infrastructure.

helpers/facebook.go
drsnyder: Ah, ok.

helpers/testhelpers/fbtesting/test_users.go
nmonterroso: The point of these are to run against a running instance of Helios, so yes we could mock them on the testing side, but they wouldn't validate on the Helios side, so the test would fail at the highest level.

We need integration tests because we currently do not have a way to automatically find errors created by [database queries](https://github.com/Wikia/helios/commit/e08ddb5756fc9556958ecac75b3ce6acc932329d#diff-630455bdc413ff1f0c6fb8561d9f3893), which are deep inside the resources being called.

nmonterroso:
We have unit tests to test out the high level interactions between Helios and Facebook in [`facebook_controller_test`](https://github.com/Wikia/helios/blob/master/controllers/facebook_controller_test.go), this is more to make sure that all the working parts that come with that (like database interactions) are working properly

drsnyder:
https://wikia-inc.atlassian.net/browse/SERVICES-668

drsnyder:
My only hesitation is not [mocking or faking](http://martinfowler.com/bliki/SelfInitializingFake.html) facebook since it's (way) out of our control. Practically this is hard because we don't have any tooling to make it easy.

nmonterroso:
Yeah, like I said earlier it would be nice if we could tell Helios via-config "use this as a Facebook gateway", but we currently don't :(. 

That also wouldn't save us from having code that just wouldn't work with the remote service. Since the disconnect functionality didn't work when tested, I can only assume that the code was never run during development, so if we had mocked everything, we'd have tests that worked, but a service that didn't, so I think this is sort of testing is (for now, at least) valuable to us.

drsnyder:
> Yeah, like I said earlier it would be nice if we could tell Helios via-config "use this as a Facebook gateway", but we currently don't have a way to do that :(.

Should we add that? It would be nice to say use-this-driver.

> That also wouldn't save us from having code that just wouldn't work with the remote service. Since the disconnect functionality didn't work when tested, I can only assume that the code was never run during development, so if we had mocked everything, we'd have tests that worked, but a service that didn't, so I think this is sort of testing is (for now, at least) valuable to us.

I think that's a different problem. Don't get me wrong, I think the test as-is is import. But I think the rate of change of the facebook API and the ownership are outside of our control. We are now breaking the "airplane" test twice if I recall correctly because now we have a dependency on the UP service and facebook.

By airplane test I mean you could install and run locally everything you need to run the tests. If we had fakes for the UP service and facebook we would still be airplane-test-able.

nmonterroso:
> Should we add that? It would be nice to say use-this-driver.
Yeah, that'd be awesome. I created https://wikia-inc.atlassian.net/browse/SERVICES-685 to track this.

I think the problem I described is the problem I'm trying to fix in this PR, and the problem you're describing (airplane test) would be helped by SERVICES-685.

drsnyder:
dope

PR #107, opened 2015-09-01
Redis client update

ArturKlajnerok:
While working on multiple token support with per user limit, I've noticed that our redis client is a bit behind the latest version, so here is the update.

/cc @Wikia/services-team


drsnyder:
Looks like failure in the build was:

```
Encountered error, last error was: Error testing server
Process leaked file descriptors. See http://wiki.jenkins-ci.org/display/JENKINS/Spawning+processes+from+build for more information
```

ArturKlajnerok:
Yeah, I'm currently checking that.

ArturKlajnerok:
@drsnyder `redigo/redis` client introduced dependency on `redigo/internal`, I've added it in : 76b5c57.
This also surfaced the problem with using standalone `godep update github.com/garyburd/redigo/redis` that don't care about new dependencies, it will update https://github.com/Wikia/helios/blob/master/README.md with proper guide for doing dependency updates in separate PR.

drsnyder:
Ok, it looks like there's no functionality change. So if the tests pass.... :sheep: it.

PR #108, opened 2015-09-02
[SERVICES-680] Support multiple tokens per user with limit.

ArturKlajnerok:
As we are moving away from `one token per user` policy, we need to have control over amount of tokens that are generated and stored per user.  For that a `user -> token list` mapping is created to keep the tokens below provided maximum.

 * [x] Remove old `user -> token` mapping
 * [x] Remove handling of single token per user
 * [x] Methods for supporting Redis lists operations
 * [x] Updating  `user -> token list` when new `access_token` is created 
 * [x] Removal of old `access_token` when max per user is reached

/cc @Wikia/services-team


nmonterroso:
This will effectively force everyone to re-log in right? That's probably ok, I just wanted to be sure. Are we orphaning all those keys in Redis?

Also, if I already have the correct cookie set and I make another request to `/token`, what happens? does my new token overwrite my old? are there two tokens? I'm not sure how browsers/clients handle this, so that info would be useful.

nmonterroso:
The implementation doesn't look like an LRU cache. The behavior looks like it's doing "when you reach the max, delete the oldest". Don't know if we really want it to be an LRU cache or if this behavior is OK, but we should be sure about our intentions.

ArturKlajnerok:
@nmonterroso 
This change will not force everyone to re-log, all the `access_tokens` will still be valid, only items with `userId -> access_token` map will become useless and they were used to support one token pre user policy. I'm also preparing a script to migrate all `userId -> access_token` mappings  to `userId -> access_token list` mappings used for limiting multiple tokens per user.

If you have the a correct cookie set and made another request to `/token`, your cookie will be overridden with the new one with new `access_token`. Your old `access_token` will still be valid until you invalidate it with `DELETE /token/{token}` or it will be invalidate when pushed out of the user token list when the max limit will be reached.

ArturKlajnerok:
@nmonterroso 
As for LRU I was considering that originally, but while doing some early R&D noticed that would require to have more complex toke structure stored in Redis and additional actions for marking lately used tokens to move them up in LRU. That would reduce the performance for validating each token that should be top priority and is responsible for 99% of traffic, so I intently changed the approach. Let me know if this solutions meets our need or should I improve it for some cases.

storage/redis_storage.go
```diff
@@ -231,24 +232,21 @@ func (storage *RedisStorage) RemoveRefresh(token string) error {
 	return storage.DeleteKey(key)
 }
 
-func (storage *RedisStorage) SaveAccessTokenForUserId(data *osin.AccessData) error {
-	key := storage.createUserIdAccessKey(data.UserData.(string))
-	return storage.SetExpirableKey(key, []byte(data.AccessToken), int32(data.ExpiresIn))
-}
+func (storage *RedisStorage) UpdateUserAccessList(data *osin.AccessData) error {
```
drsnyder: Can we take this opportunity to decouple this part of the code from `osin`?

drsnyder:
What will happen if you have 10 active tokens and you login one more time? Will you be force logged out of your oldest session?

drsnyder:
Can `UserAccessListMaxSize` be configurable? AFAIK we don't have any data on active sessions.

drsnyder:
We should consider using redis sorted sets for this instead of lists. If we did that we would get two benefits. The first is that it ensures all `UserAccessListMaxSize` tokens are unique. It would also make implementing an LRU fairly trivial.

Regarding the LRU, the "score" used with the `Z*` methods could be used to track the usage rate. Whenever a token is validated you could update the score with `ZINCRBY`. When you need to remove one you can remove the first one from `ZRANGE`.

nmonterroso:
@drsnyder do you smell that?

drsnyder:
image

storage/redis_storage.go
ArturKlajnerok: I've removed `osin` reference from that method : ebec4a4

ArturKlajnerok:
@drsnyder Moved `UserAccessListMaxSize` to config : a1f3506

ArturKlajnerok:
@drsnyder If you will have max tokens already and login again the oldest one will be removed and you will be effectively logged out from oldest session.

storage/redis_storage.go
```diff
@@ -231,24 +233,21 @@ func (storage *RedisStorage) RemoveRefresh(token string) error {
...
 
-func (storage *RedisStorage) GetAccessForUserId(userId string) (*osin.AccessData, error) {
-	key := storage.createUserIdAccessKey(userId)
-	accessToken, err := storage.GetKey(key, false)
-	if err != nil {
-		return nil, err
-	}
-	if accessToken == nil {
-		return nil, nil
+	if err == nil && length > storage.maxTokensPerUser {
```
nmonterroso: Should this be behind a getter that tries to read the config, but will fall back to a default (10? 25?) if the config isn't present? Since we have to release the code before the chef config, for a while in production this won't be set. We should make sure that this still works in this case.

drsnyder: I like the idea of having some kind of config interface. That might also make it easier for us to swap out the current config file implementation.

storage/redis_storage.go
```diff
@@ -231,24 +233,21 @@ func (storage *RedisStorage) RemoveRefresh(token string) error {
...
+		token, err := storage.RemoveLastFromList(key)
+		if err == nil {
+			err = storage.RemoveAccess(token)
+		}
 	}
```
harnash: Would be nice to move this functionality (size limited stack list operations) to a separate function.

storage/redis_storage.go
```diff
@@ -305,11 +304,47 @@ func (storage *RedisStorage) DeleteKey(keyName string) error {
 
 	db := pool.Get()
 	defer db.Close()
-	_, err = db.Do("DEL", keyName)
+	_, err = db.Do("DEL", key)
 	logger.GetLogger().ErrorErr(err)
 	return err
 }
 
+func (storage *RedisStorage) AddToList(key string, value []byte) (int64, error) {
```
harnash: I would follow REDIS convention using PUSH and POP keywords. They are more verbose and reveals the mechanisms behind them.

pchojnacki: +1

harnash:
I think we missed the point. The goal was to have unique tokens *per application* here we allow user to generate up to X tokens every time he authorises and we drop access tokens when limit is reached. Limiting tokens is OK but we don't have information about the token<->application association. If user log's in into 10+ different places (mobile apps, mobile browser, desktop etc.) it will be logged out if he exceeds the limit. This violates the rule of least surprise. Can we do something about that?

storage/redis_storage.go
nmonterroso: Hm, actually I think it would be better to create an interface

```go
type AccessTokenStorage interface {
  ...
}

type RedisAccessTokenStorage struct {
  ...
}
```

nmonterroso:
@harnash i don't think we can. There's no way to differentiate a login request from MediaWiki from a login request from Mercury or any given mobile app. I think this was the original idea with having client id/secret pairs for each app, but since we don't support that anymore we have to go this route.

storage/redis_storage.go
nmonterroso: @drsnyder that seems like a good idea, but would require a rework in many places. I've created [a ticket](https://wikia-inc.atlassian.net/browse/SERVICES-706) to track this.

For now, I've put the config behind a getter that will return a sane default in 90467e3

nmonterroso:
@drsnyder @ArturKlajnerok 
I've updated the implementation to use an LRU cache backed by redis sorted sets in 8b7b54f. Please take a look.

nmonterroso:
We probably can't merge this until https://wikia-inc.atlassian.net/browse/SERVICES-692 is completed. We should probably make those changes in a branch off of this branch, and then merge those changes together to master

storage/redis_storage.go
```diff
@@ -310,11 +324,135 @@ func (storage *RedisStorage) DeleteKey(keyName string) error {
...
+
+	if err != nil {
+		logger.GetLogger().ErrorErr(err)
+		return false
+	}
+
+	now := getCurrentTimestamp()
+	return now - lastUpdated > storage.accessTokenUpdateInSec
+}
+
```
drsnyder: Can we add integration tests for this functionality to make sure it behaves as we expect?

harnash:
@nmonterroso I don't agree. We can change it if we really want to. We could do this in two phases: support client_id and client_secret if they are specified and if not fall back to hardcoded one until we migrate all the clients to use proper OAuth mechanism. By hacking around this library we probably end adding similar solutions as this one to problems which would not be an issue if we did stick to OAuth spec from the beginning. I bet it will not be long enough when limit of 10 tokens will be to low (how many mobile apps we have now?) and we get issues that some users get logged out constantly because they logged in into more than 10 apps. It will bite us sooner or later.
I don't know why we decided to go this route and I think we still can go back since there are relatively few clients using this API.

nmonterroso:
@harnash we intentionally removed support for client id/secret a while ago as part of [this ticket](https://wikia-inc.atlassian.net/browse/SERVICES-361). If we want to change that we should talk about it as a team and plan accordingly. 

WRT how many tokens each user is allowed to have, 10 is only the default, and it can be updated via config

harnash:
@nmonterroso Yeah. We should discuss this since it raises questions. We could release this as temporary fix for 'one token per user' case but we should be weary that it might hit users with many apps.

ArturKlajnerok:
@harnash The original approach was based on `Oauth2` that was implementing properly supporting for multiple users and clients. But after architectural recommendation we started to move to less complex `token` approach and all the changes that were lately introduced are consequence of that. Also in this case we are full aware of implications connected with those changes. Let's clarify that more directly as it seems that there is some miscommunication in that matter.

storage/redis_storage.go
```diff
@@ -381,8 +519,12 @@ func (storage *RedisStorage) createRefreshKey(token string) string {
...
 }
 
-func (storage *RedisStorage) createUserIdAccessKey(userId string) string {
-	return storage.prefix + UserIdAccessKeyPrefix + userId
+func (storage *RedisStorage) createUserAccessListKey(userId string) string {
+	return storage.prefix + UserAccessSortedSetPrefix + userId
+}
+
+func getCurrentTimestamp() int64 {
+	return time.Now().Unix()
```
ArturKlajnerok: Should we use `UTC()` ?

ArturKlajnerok:
Lgtm :+1:

PR #109, opened 2015-09-02
changing /info http method to GET

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-462

We changed the HTTP method of `/info` to `POST`, but the api gateway and MediaWiki currently use `GET`, and that also seems more appropriate, so changing this back to `GET /info`


ArturKlajnerok:
Seems that e2e test for token invalidation are failing.

nmonterroso:
@ArturKlajnerok yah i'm looking.

nmonterroso:
looks like `code` is no longer being accepted as a form param in the `get` request, which seems correct, but I wonder why it was working before O.o

ArturKlajnerok:
Lgtm :+1:

PR #110, opened 2015-09-02
clear out mediawiki session cookie

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-677

also clear out the mediawiki session cookie when logging out of helios


drsnyder:
lgtm

PR #112, opened 2015-09-04
[SERVICES-699] Removing refresh token flow and storage.

ArturKlajnerok:
Since we are using long living `access_token` and not using actively `refresh_token`, support for it only generates complexity and additional cost for changes and new futures. In addition `refresh_token` won't waste our storage space anymore.

/cc @Wikia/services-team


models/user.go
```diff
@@ -26,7 +26,6 @@ const (
 	AuthenticateFailed         = iota
 	AuthenticatedNormalPass    = iota
 	AuthenticatedTempPass      = iota
-	AuthenticatedRefreshToken  = iota
```
nmonterroso: This will change all the constants. I wonder if anyone is depending on their current values (I know MW is using the current value of `AuthenticatedTempPass`), so maybe we should change these to be their actual values instead of using `iota`

nmonterroso:
Since this changes the output of requests that return a token payload to no longer include the refresh token, we should communicate to other teams that they will no longer be a part of the response.

models/user.go
ArturKlajnerok: The constants that are after `AuthenticatedRefreshToken` were introduced lately so I don't think those values are used, but I will double check if no client is using them. Considering that `iota` should be still ok.

ArturKlajnerok:
@nmonterroso The output will not change, this PR only removes support for storing and using `refresh_token` as generating them as part of `access_token` is still in control of external library. I will still communicate that change to be sure that on active client uses the `refresh_token. :+1:

harnash:
Code looks good but I would prefer not to send `refresh_token` in responses. At least send only empty string not to confuse client that this value may be used for authentication or anything.

ArturKlajnerok:
@harnash We don't have much control over `refresh_token` generation that is handled with `/token` resource via the `osin` library and removing it from the response at the end of the process is  also not a clean solution, so I would leave it and no client have a way to use it after those changes. In general we moved away form `Oauth2` approach to standard `token` approach, so at some point we could remove the `osin` dependency and ale the artefacts connected.

harnash:
@ArturKlajnerok I know we can't probably remove refresh token generation completely since it is done inside osin but we could clear it just before sending response somewhere around https://github.com/Wikia/helios/blob/master/controllers/oauth_controller.go#L88 and alike. Returning partially invalid values also is not a clean solution and confusing.
If that is a big problem - skip it.

harnash:
I'm going to merge it but we should special attention when releasing this so we test it in preview and monitor it on prod after release.

PR #113, opened 2015-09-08
SERVICES-710 invalidate token responses

harnash:
This will fix the response status codes when invalidating user access tokens and improve logging.

@Wikia/services-team


controllers/oauth_controller.go
```diff
@@ -213,17 +214,28 @@ func (controller *OAuthController) tokenInvalidateHandler(w http.ResponseWriter,
...
+		logger.GetLogger().Error(fmt.Sprintf("Could not fetch token for user %v using token: '%v'", requestingUserId, token))
+		controller.setOutputError(resp, osin.E_SERVER_ERROR)
+		resp.StatusCode = http.StatusInternalServerError
+	} else if errors.ContainsError(service.AuthenticationError) {
+		logger.GetLogger().Warn(fmt.Sprintf("Could not authenticate user %v for token: '%v'", requestingUserId, token))
+		controller.setOutputError(resp, osin.E_UNAUTHORIZED_CLIENT)
+		resp.StatusCode = http.StatusUnauthorized
+	} else {
+		invalidateAccessTokenCookie(resp, w, controller.config.Server.Domain)
+		invalidateMediaWikiSessionCookie(resp, w, controller.config.Server.Domain)
```
sqreek: Maybe we should send token that was deleted?

harnash: We could but currently I don't see the use case for that and it will increase bandwidth and latency. You still have the token in URI so duplicating that info seems unnecessary. If we needed to send some more info about deleted token then maybe but currently I would say we should stay with no content.

sqreek: So we will logout user from mediawiki, if user want to logout from any other application?

sqreek:
LGTM

PR #114, opened 2015-09-08
adding auth to logout swagger spec

nmonterroso:
@Wikia/services-team 
updating the swagger spec for logout to indicate that it requires authentication


ArturKlajnerok:
Lgtm :+1:

PR #115, opened 2015-09-09
Revert "[SERVICES-699] Removing refresh token flow and storage."

harnash:
Reverts Wikia/helios#112

We just got confirmation that mobile apps do use refres_token so I'm going to revert this merge for now.


sqreek:
LGTM

aniaMu:
@harnash @sqreek we are working on removing that field, so I'll let you know after we do that. Should be ready today.

PR #117, opened 2015-09-09
Disabling token update

ArturKlajnerok:


harnash:
@ArturKlajnerok Build failed please revisit the change

PR #118, opened 2015-09-10
Revert "Revert "[SERVICES-699] Removing refresh token flow and storage.""

ArturKlajnerok:
We have confirmation from all clients that `refresh_token` support can be dropped. So I'm reverting the temporary revert of those changes.

This reverts commit e911861f93f7133396351fdd6a7c670aff656543.

/cc @Wikia/services-team


nmonterroso:
LGTM shipit

PR #119, opened 2015-09-17
[SERVICES-776] Added accept application/json header when setting preferences

ArturKlajnerok:
We need to support `Accept: application/json` header for `PUT user-preference/{userId}`.

/cc @Wikia/services-team


nmonterroso:
LGTM shipit!

PR #120, opened 2015-09-21
SERVICES-742: Added support for Facebook v.1.x lined users. [DO NOT MERGE]

harnash:
On Media Wiki side there are accounts whose fb_app_id is set to 0. Those
accounts were linked before migrating to new 2.x API. We need to support
them until they are fully migrated.


harnash:
We need to clarify if we are going to support legacy users in Helios. If we do we need to allow Helios to properly lookup them in the MW DB.

ArturKlajnerok:
@harnash I think there will still be problem with `fb access token` validation, as we are providing `fb app id` and `fb app secret` when using fb client. https://github.com/Wikia/helios/blob/5092d2f601bd2bd5af954cc6515f95047bec3558/helpers/facebook.go#L5

So I think with this strict app usage the `fb access token` for old fb accounts with `fb_app_id = 0` will be not be validated properly.

harnash:
@gbenson-wikia You are right. Fixed that in ecf29e7.

harnash:
@ArturKlajnerok You might be right but what I saw is that I got token from legacy user and Graph did validated it properly and it seemed that user lookup failed. I need this change (even not merged) to properly test this case on prod using off-site tool I wrote.

ArturKlajnerok:
I've tested today updating `fb_app_id` to our main app id for facebook v1 user, it came out that after that everything worked properly. I will proceed with the update and we can close this PR.

PR #121, opened 2015-09-25
Create an auth interface for purging all of a users access tokens

drsnyder:
This PR creates an `auth` interface with `PurgeAllUserAccessTokens` that when called should effectively log the user out of all of their sessions. This is accomplished by deleting their session token set and all of the keys associated with their tokens.

Along with this PR, I'm adding an interface to the redis storage implementation so that its functionality can be mocked in the auth service tests. With this interface we might want to also change the call signatures of some of the auth methods since they are currently accepting a concrete implementation of the redis client as a parameter. I'm going to capture that in another ticket.

https://wikia-inc.atlassian.net/browse/SERVICES-826

TODO
* [x] integration tests for the new redis command
* [x] fix the initialization of the auth service in the controllers


nmonterroso:
I think that this is a good step in moving us away from `osin`, since potentially we could move away the methods `LoadAccess`, `RemoveAccess`, and other related methods currently in `RedisStorage` into the `AuthTokenStorage` interface.

service/auth.go
```diff
@@ -18,12 +19,15 @@ type AuthService interface {
 	DeleteAccessToken(redisStorage *storage.RedisStorage, token string, requestUserId int64) (bool, *errorlist.ErrorList)
```
nmonterroso: I wanted to suggest moving this function into `AccessTokenStorage` as 

```go
DeleteAccessToken(token string, requestUserId int64) (bool, *errorlist.ErrorList)
```

but `DeleteAccessToken` relies on the internals of the `osin` package to do the deletion, so can't do that yet :(

drsnyder: Yeah, the `osin` dependency seems out of place in `storage`.

drsnyder: Oh, well we can do that now for the sake of dropping the first parameter and using the injected `RedisStorage`. We just have to add the methods used in `DeleteAccessToken` to the `AccessTokenStorage`. I created SERVICES-830.

storage/redis_storage.go
```diff
@@ -172,30 +176,34 @@ func (storage *RedisStorage) RemoveAuthorize(code string) error {
...
 }
 
+// this is being preserved for osin compatability
 func (storage *RedisStorage) SaveAccess(data *osin.AccessData) error {
-	key := storage.createAccessKey(data.AccessToken)
+	return storage.SaveAccessToken(data.UserData.(string), data.AccessToken,
+		int32(data.ExpiresIn), data)
+}
+
+func (storage *RedisStorage) SaveAccessToken(userId, token string, expires int32, data interface{}) error {
```
drsnyder: I created this so that the tests are not dependent upon the `osin` data structures.

storage/redis_storage.go
```diff
@@ -209,6 +217,20 @@ func (storage *RedisStorage) LoadAccess(token string) (*osin.AccessData, error)
 	return accessData, err
 }
 
+func (storage *RedisStorage) GetTokenData(token string) ([]byte, error) {
```
drsnyder: I pulled this out from `LoadAccess` also for the sake of decoupling it the `osin` data structures and making testing easier.

storage/redis_storage.go
```diff
@@ -311,7 +410,7 @@ func (storage *RedisStorage) SetExpirableKey(key string, value []byte, expireInS
 }
 
 func (storage *RedisStorage) DeleteKey(key string) error {
-	pool, err := storage.getPoolForRead()
+	pool, err := storage.getPoolForWrite()
```
drsnyder: Using the "read" pool didn't seem consistent with delete.

storage/redis_storage.go
```diff
@@ -261,6 +283,85 @@ func (storage *RedisStorage) UpdateUserAccessSortedSet(userId, token string) err
...
+	_, err = db.Do("ZREMRANGEBYRANK", key, 0, -1)
+	if err != nil {
+		logger.GetLogger().ErrorErr(err)
+		return err
+	}
+
+	// delete all of the individual tokens
+	for _, token := range tokenList {
+		tokenKey := storage.createAccessKey(token)
+		_, err = db.Do("DEL", tokenKey)
```
harnash: We should log what we delete and why

service/auth.go
```diff
@@ -81,3 +85,16 @@ func (service *AuthServiceValue) DeleteAccessToken(redisStorage *storage.RedisSt
 
 	return true, errors
 }
+
+func (service *AuthServiceValue) LogoutUserFromAllSessions(userId int64) (bool, *errorlist.ErrorList) {
+	errors := errorlist.NewErrorList()
+
+	err := service.accessTokenStorage.PurgeAllUserAccessTokens(strconv.FormatInt(userId, 10))
+	if err != nil {
+		logger.GetLogger().Error(fmt.Sprintf("Error logging user id '%v' out from all sessions: %v", userId, err))
```
sqreek: `Error logging out from all sessions for user '%v': '%v'` ?

service/auth.go
```diff
@@ -81,3 +85,16 @@ func (service *AuthServiceValue) DeleteAccessToken(redisStorage *storage.RedisSt
...
+		errors.Add(errorlist.GenericDataAccessError)
+		return false, errors
+	}
+
+	return true, errors
```
sqreek: `return true, nil`?

harnash:
LGTM (except logging)

storage/redis_storage.go
drsnyder: 46353e5

service/auth.go
drsnyder: Our convention has been to return an empty error list instead of `nil`. That way, you can confirm that the list is empty with `errors.IsEmpty()`.

nmonterroso:
:shipit: real good

storage/redis_storage.go
ArturKlajnerok: Maybe `LoadTokenData` would be more consistent with naming methods in storage ?

storage/redis_storage.go
```diff
@@ -261,6 +283,86 @@ func (storage *RedisStorage) UpdateUserAccessSortedSet(userId, token string) err
...
+		return err
+	}
+
+	return nil
+}
+
+func (storage *RedisStorage) GetUserSortedSet(userId string) ([]string, error) {
+	tokenKeyList := make([]string, 0)
+	key := storage.createUserAccessSortedSetKey(userId)
+	pool, err := storage.getPoolForWrite()
```
ArturKlajnerok: I thin we should use here `getPoolForRead()`.

storage/redis_storage.go
drsnyder: I'm used to seeing read operations on the read pool and write operations on the write pool. Should the reasoning about which to use be different here or am I misunderstanding the meaning of `getPoolForRead`/`getPoolForWrite`?

nmonterroso: I think you have the right idea, though in practice there are no consequences to which one you choose, since the configs for redis read/write instances are identical.

ArturKlajnerok: `read operations on the read pool and write operations on the write pool` this is correct, and delete also should be done on `getPoolForWrite()` so this is a good catch. As far I remember  @jcellary  introduced that to have a possibility to switch to slave when master if master is down with `forceUseSlave` config flag. But for now both methods are using master instance, I think we should change that and serve all reads from slave.

storage/redis_storage.go
drsnyder: c3a09ec

storage/redis_storage.go
ArturKlajnerok: I've addressed that in SERVICES-839.

PR #122, opened 2015-09-28
Introduce RequestAuthenticator

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-825

introducing RequestAuthenticator which can authenticate requests, either with a schwartz token or  `X-Wikia-UserId` header. Also updated all the relevant tests/controller methods to use this.

TODO:
- [x] tests for `RequestAuthenticatorValue`


helpers/request_authenticator.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func NewRequestAuthenticator(schwartzToken string) RequestAuthenticator {
+	return &RequestAuthenticatorValue{schwartzToken}
+}
+
+func (authenticator *RequestAuthenticatorValue) SchwartzAuthenticate(request *http.Request) bool {
+	token := request.Header.Get(SchwartzTokenHeaderName)
+	return token == authenticator.schwartzToken
+}
```
drsnyder: do you have the Schwartz?

helpers/request_authenticator.go
```diff
@@ -0,0 +1,39 @@
...
+
+func (authenticator *RequestAuthenticatorValue) AuthorizeAction(request *http.Request, targetUserId int64) bool {
+	authenticatedId := authenticator.GetAuthorizedUser(request)
+	return authenticatedId != 0 && authenticatedId == targetUserId
+}
+
+func (authenticator *RequestAuthenticatorValue) GetAuthorizedUser(request *http.Request) int64 {
+	return utils.ParseInt64(request.Header.Get(common.GatewayUserIdHeader))
```
drsnyder: What happens here if the value is `nil`?

drsnyder:
Minor comments, OTW, LGTM

helpers/request_authenticator.go
nmonterroso: a value from `request.Header.Get` will return an empty string if the header is not present, and `utils.ParseInt64` will return `0` on that. I'll add a test so there's no confusion.

nmonterroso: c13bceb

PR #123, opened 2015-09-28
Use getPoolForRead when getting the sorted set

drsnyder:
I mistakenly missed this change in PR #121. Fixing it here.

/cc @ArturKlajnerok


ArturKlajnerok:
Lgtm :+1:

PR #126, opened 2015-09-30
Expand AccessTokenStorage

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-830

Moving the functionality used by `AuthService.DeleteAccessToken` into `AccessTokenStorage`.


ArturKlajnerok:
If we are refactoring `DeleteAccessToken`, I think it would be also good to remove token key and entry in the set as a transaction. Similar transaction is done for multiple token removal : https://github.com/Wikia/helios/blob/master/storage/redis_storage.go#L307

nmonterroso:
@ArturKlajnerok Good idea. I did this in fc649a70d1a912ef275582af9db7d33f6628f73a and also used other transactions to use the same mechanism

service/auth.go
```diff
@@ -46,39 +46,31 @@ func (service *AuthServiceValue) GenerateTemporaryPassword(length int) (string,
 	return password, errors
 }
 
-func (service *AuthServiceValue) DeleteAccessToken(redisStorage *storage.RedisStorage, token string, requestUserId int64) (bool, *errorlist.ErrorList) {
+func (service *AuthServiceValue) DeleteAccessToken(token string, requestUserId int64) (bool, *errorlist.ErrorList) {
 	errors := errorlist.NewErrorList()
-	accessData, err := redisStorage.LoadAccess(token)
+	userId, err := service.accessTokenStorage.GetUserIdFromToken(token)
```
drsnyder: :+1:

service/auth.go
```diff
@@ -46,39 +46,31 @@ func (service *AuthServiceValue) GenerateTemporaryPassword(length int) (string,
...
 	logger.GetLogger().Debug(fmt.Sprintf("Deleting token for user: %v", userId))
-	err = redisStorage.RemoveAccessFromUserSet(userId, token)
-	if err != nil {
-		logger.GetLogger().Error(fmt.Sprintf("Error deleting mapping user -> token (access token '%v'): %v", token, err))
-		errors.Add(errorlist.GenericDataAccessError)
-		return false, errors
-	}
-
-	err = redisStorage.RemoveAccess(token)
+	err = service.accessTokenStorage.RemoveAccessFromUserSet(userId, token)
```
drsnyder: Minor: `RemoveAccessFromUserSet` exposes some of the underlying implementation details. That might be ok... but maybe `RemoveAccess` is ok or `RevokeAccessToken(token)` is better?

storage/redis_storage.go
```diff
@@ -470,26 +470,9 @@ func (storage *RedisStorage) pruneSortedSetToSize(key string, maxSize int64) ([]
...
-		return removedElements, err
-	}
-	response, err := db.Do("EXEC")
-	if err != nil {
-		logger.GetLogger().ErrorErr(err)
-		return removedElements, err
-	}
+	response, err := storage.multi(
+		newTransactionStep("ZRANGE", key, removeStart, removeEnd),
+		newTransactionStep("ZREMRANGEBYRANK", key, removeStart, removeEnd))
```
drsnyder: :+1:

drsnyder:
lgtm

ArturKlajnerok:
@nmonterroso Nice one :+1:

service/auth.go
nmonterroso: That's a good point, we'd never want to remove an access token without removing it from the set, or remove from the set without removing the access token, so they should always be done together. The right function name though is probably `RemoveAccess`, since that's part of the `osin.Storage` interface

nmonterroso: I changed this function to be named `RemoveAccessTokenForUser(userId, token)` and changed `RemoveAccess` to call this function in 7981e9f

service/auth.go
```diff
@@ -46,39 +46,31 @@ func (service *AuthServiceValue) GenerateTemporaryPassword(length int) (string,
...
 	if err != nil {
-		logger.GetLogger().Error(fmt.Sprintf("Error deleting access token (access for token '%v'): %v", token, err))
```
ArturKlajnerok: We should also log here the `err` message.

PR #127, opened 2015-10-02
Add auth service interface to check the username and IP address against phalanx for blocks

drsnyder:
This PR adds `auth.IsUserBlocked(userId, ipAddress)`. Internally, it looks up the user in the database, then calls phalanx to check the username and the ipaddress.

As part of this PR, I also added a `SystemContainer` interface and wired it into the auth service to help combat the telescoping constructor.

https://wikia-inc.atlassian.net/browse/SERVICES-852

@Wikia/services-team 

TODO:
 * [x] fix any breakage from changing the auth service creation
 * [x] ~~determine if we need to parallelize the phalanx requests and make the change~~ (let's do that in a different PR)


service/auth.go
```diff
@@ -20,21 +20,23 @@ type AuthService interface {
 	// this is mostly a wrapper that can convert an error to an ErrorList and log
 	GenerateTemporaryPassword(length int) (string, *errorlist.ErrorList)
 	LogoutUserFromAllSessions(userId int64) *errorlist.ErrorList
+
+	IsUserBlocked(userId, ipAddress string) (bool, *errorlist.ErrorList)
 }
 
-type AuthServiceValue struct {
-	accessTokenStorage storage.AccessTokenStorage
+type authServiceValue struct {
```
drsnyder: I made this private. As far as I know, there shouldn't be any reason to export these `*Value` structs.

service/auth.go
```diff
@@ -46,9 +48,9 @@ func (service *AuthServiceValue) GenerateTemporaryPassword(length int) (string,
 	return password, errors
 }
 
-func (service *AuthServiceValue) DeleteAccessToken(token string, requestUserId int64) (bool, *errorlist.ErrorList) {
+func (service *authServiceValue) DeleteAccessToken(token string, requestUserId int64) (bool, *errorlist.ErrorList) {
 	errors := errorlist.NewErrorList()
-	userId, err := service.accessTokenStorage.GetUserIdFromToken(token)
+	userId, err := service.system.GetAccessTokenStorage().GetUserIdFromToken(token)
```
drsnyder: This breaks the law of Demeter but I think it's ok in this case since `system` is the DI container.

service/auth.go
```diff
@@ -89,3 +91,42 @@ func (service *AuthServiceValue) LogoutUserFromAllSessions(userId int64) *errorl
 
 	return errors
 }
+
+func (service *authServiceValue) IsUserBlocked(userId, ipAddress string) (bool, *errorlist.ErrorList) {
+	errors := errorlist.NewErrorList()
+
+	userIdInt, _ := strconv.ParseInt(userId, 10, 64)
+	user, err := service.system.GetUserDAO().LookupById(userIdInt)
```
drsnyder: We should consider caching this but we'll have to be careful. We probably don't want to cache the whole user object.

As far as I know we haven't added any ephemeral caching to helios yet. I'm not familiar with whether or not we change user names so depending on how that works any caching we do here would have to take that into account.

service/auth.go
```diff
@@ -89,3 +91,42 @@ func (service *AuthServiceValue) LogoutUserFromAllSessions(userId int64) *errorl
...
+
+	return service.areAnyStringsBlocked(user.Name, ipAddress)
+}
+
+func (service *authServiceValue) areAnyStringsBlocked(args ...string) (bool, *errorlist.ErrorList) {
+	errors := errorlist.NewErrorList()
+
+	for _, stringToCheck := range args {
+		// returns true when the string is *not* blocked
+		ok, err := service.system.GetPhalanxClient().CheckName(stringToCheck)
```
drsnyder: We should do this concurrently. I want to get it "right" first. Perhaps the caching mentioned above and the concurrency could be follow on PRs so they don't distract from the business logic we are adding.

service/user_test.go
```diff
@@ -229,14 +221,6 @@ func (suite *UserServiceTestSuite) TestSendPasswordResetEmailSuccess() {
 	assert.True(suite.T(), errors.IsEmpty())
 }
 
-func getBillAsModel() *models.User {
```
nmonterroso: poor Bill :(

nmonterroso: Oh nevermind he lives somewhere else

service/auth.go
nmonterroso: yah, since they're instantiated through the `NewThing` pattern this is probably a good change :+1:

service/auth.go
nmonterroso: There is a `UserRenameTask` in mediawiki

nmonterroso:
LGTM

PR #128, opened 2015-10-03
[SERVICES-853] Checking for user/ip block on token validation resource

ArturKlajnerok:
Each time token will be validated addition check for username or ip block will be performed. This will be default behavior for `/info` resource and can be turned off with additional param `noblockcheck`. For blocked request `403 Forbidden` will be returned.

* [x] Check for user/ip block
* [x] Off switch with `noblockcheck` param
* [x] integration tests

/cc @Wikia/services-team


controllers/oauth_controller.go
```diff
@@ -83,11 +88,39 @@ func (controller *OAuthController) GetRedisStorage() *storage.RedisStorage {
...
+}
+
 func (controller *OAuthController) infoHandler(w http.ResponseWriter, r *http.Request) {
 	resp := controller.server.NewResponse()
 	defer resp.Close()
+	defer osin.OutputJSON(resp, w, r)
 
 	if ir := controller.server.HandleInfoRequest(resp, r); ir != nil {
+		r.ParseForm()
+		noBlockCheck := r.Form["noblockcheck"]
```
nmonterroso: does `r.Form` also contain query parameters?

service/auth.go
```diff
@@ -106,7 +106,16 @@ func (service *authServiceValue) IsUserBlocked(userId, ipAddress string) (bool,
 		return false, errors
 	}
 
-	return service.areAnyStringsBlocked(user.Name, ipAddress)
+	if user == nil {
+		logger.GetLogger().ErrorMap(map[string]interface{}{
```
nmonterroso: In what cases would we fall back to this? It seems like in all current usages are checking for valid users, after the user has been loaded (so they must exist).

controllers/oauth_controller.go
```diff
@@ -83,11 +88,39 @@ func (controller *OAuthController) GetRedisStorage() *storage.RedisStorage {
 	return controller.redisStorage
 }
 
+func (controller OAuthController) checkBlockedUser(resp *osin.Response, userRemoteAddres string, userId string) bool {
```
drsnyder: `userRemoteAddress`?

drsnyder: Can we decouple this from `osin`? Also, please add tests. It's not clear to me why we are splitting the `userRemoteAddres`.

controllers/oauth_controller.go
```diff
@@ -83,11 +88,39 @@ func (controller *OAuthController) GetRedisStorage() *storage.RedisStorage {
...
+		setRequestError(resp, UserBlockedTitle, UserField, UserBlocked)
+		resp.StatusCode = http.StatusForbidden
+	} else if errors.ContainsError(errorlist.GenericDataAccessError) {
+		logger.GetLogger().Error(fmt.Sprintf("Could not check if user '%v' is blocked", userId))
+		resp.StatusCode = http.StatusInternalServerError
+		blocked = true
+	}
+
+	return blocked
+}
```
drsnyder: We should determine how we want to handle the case where there is an issue with phalanx. Do we let the request go through? Do we block by default?

controllers/oauth_controller.go
harnash: Yes: https://golang.org/src/net/http/request.go:
```
// Form contains the parsed form data, including both the URL
// field's query parameters and the POST or PUT form data.
// This field is only available after ParseForm is called.
// The HTTP client ignores Form and uses Body instead.
```

service/auth.go
harnash: This was failing in tests when we tried to authenticate not existent user. This shouldn't happen on prod but I think we should pay more attention to some edge cases.

controllers/oauth_controller.go
harnash: Right. Name is not the best I think -> 1cae800
We get client remote IP and PORT so we split it into two. Another thing is that we do only support IPv6 addresses.
Also tests are on TODO list.

controllers/oauth_controller.go
harnash: Correct me if I am wrong but looking at Phalanx implementation in Wikia/app if there is an error fetching record from DB user is not blocked.

drsnyder: Ok, at a minimum, we'll be consistent with MW.

harnash:
Tests fails on Jenkins for no apparent reason. I have added user to Phalanx on prod so it should pass.

harnash:
The problem was with one Phalanx instance which had outdated config and could not connect to DB.

service/auth.go
```diff
@@ -115,6 +124,8 @@ func (service *authServiceValue) areAnyStringsBlocked(args ...string) (bool, *er
 	for _, stringToCheck := range args {
 		// returns true when the string is *not* blocked
 		ok, err := service.system.GetPhalanxClient().CheckName(stringToCheck)
+		logger.GetLogger().Debug(fmt.Sprintf("Phalanx '%v' is valid: %v", stringToCheck, ok))
```
nmonterroso: Do we need this log? We'll probably generate a lot of messages in production since this is called on the `/info` resource which has a huge amount of traffic. If this won't show up in production because of log level filtering through config then that's fine but we probably shouldn't log the success case.

nmonterroso: I just checked and looks like we have no way to initialize the log level differently in production vs dev. We do `logger.InitLogger(common.AppName, logger.LogLevelDebug)` everywhere, so we should probably remove this log line.

harnash: Yeah. That is on my todo list. I will remove this for now.

controllers/oauth_controller.go
```diff
@@ -83,11 +89,39 @@ func (controller *OAuthController) GetRedisStorage() *storage.RedisStorage {
...
+	if !utils.IsLikeIP(address[0]) {
+		logger.GetLogger().Warn(fmt.Sprintf("Address '%v' doesn't look like an ip address", address[0]))
+	}
+
+	blocked, errors := controller.authService.IsUserBlocked(userId, address[0])
+	if blocked {
+		logger.GetLogger().Info(fmt.Sprintf("User '%v' is blocked", userId))
+		setRequestError(resp, UserBlockedTitle, UserField, UserBlocked)
+		resp.StatusCode = http.StatusForbidden
+	} else if errors.ContainsError(errorlist.GenericDataAccessError) {
```
sqreek: Should we first handle errors, then check logic?
We have this convention in other places

harnash: @sqreek: fixed in e8d76ea

controllers/oauth_controller_e2e_test.go
```diff
@@ -221,6 +221,35 @@ func (suite *OauthTestSuite) TestE2eOauthInvalidateNonExistingToken() {
...
+	response := suite.callOauthEndpoint(TokenEndpoint, values, map[string]string{}, MethodPOST)
+
+	suite.testTokenResponse(response)
+	infoPath := createInfoResourcePath(response.AccessToken)
+	response = suite.callOauthEndpoint(infoPath, url.Values{}, map[string]string{}, MethodGET)
+
+	assert.Equal(suite.T(), http.StatusForbidden, response.HttpStatus, common.BadHttpStatus)
+}
+
+func (suite *OauthTestSuite) TestBlockedUserNoBlock() {
```
sqreek: What this test do?
BlockedUserNoBlock?

harnash: It check if blocked user is not blocked when we pass extra flag `noblockcheck` to skip Phanalx check.

harnash: @sqreek I have changed the name of the test to be more descriptive: 9022e7b

sqreek:
LGTM

PR #129, opened 2015-10-09
[SERVICES-908] Increasing temporary pass length.

ArturKlajnerok:
This changes the temp pass length and adds more logs info around token resource.

/cc @Wikia/services-team


drsnyder:
lgtm

controllers/oauth_controller.go
```diff
@@ -190,7 +195,11 @@ func (controller *OAuthController) tokenGenerateHandler(w http.ResponseWriter, r
 	defer resp.Close()
 	defer osin.OutputJSON(resp, w, r)
 
-	logger.GetLogger().Debug("Starting access token generation")
+	logger.GetLogger().InfoMap(map[string]interface{}{
+		"@message": "Starting access token generation",
+		"userName": r.Form.Get(common.UsernameField),
+		"ip":       r.Header.Get(controller.config.Server.IpHeader),
+	})
```
drsnyder: This log might be unnecessary. I think if we log this once it's ok.

harnash: @drsnyder Can you point the place where is the duplicate of this log?

drsnyder: I was referring to the [one on line 214 below](https://github.com/Wikia/helios/pull/129/files#diff-94fddb1d6c9926065dc08b9e25fb0205R214). It's minor, but this line and the one on 214 are nearly identical. We'll probably be more interested in the one below.

PR #130, opened 2015-10-09
[SERVICES-886] Pass user language to email confirmation 

ArturKlajnerok:
We are forced to pass the user language code directly to the confirmation email, to avoid a race condition when using preferences.

/cc @Wikia/services-team


service/user.go
```diff
@@ -191,7 +191,8 @@ func (service *UserServiceValue) SendConfirmationEmailCommand(user domain.User)
 		return true, nil
 	}
 
-	if send, err := service.wikiaClient.SendConfirmationEmail(user.Name()); !send {
+	langPreference := user.Preferences().GetByName(common.LanguageField)
+	if send, err := service.wikiaClient.SendConfirmationEmail(user.Name(), langPreference.Value); !send {
```
nmonterroso: This will return `""` on a non-existent preference (like when the user doesn't specify that in the request)?

nmonterroso:
minor comment otw :ship:

service/user.go
ArturKlajnerok: That is correct, `langPreference.Value` will have value `""` and then default `en` email will be send. But this will not happen in current setup as we are also setting default when form param is empty : https://github.com/Wikia/helios/blob/master/controllers/helpers.go#L104

nmonterroso: Cool, just wanted to make sure it wasn't going to blow up if someone didn't provide that `langCode` param. :shipit:

PR #132, opened 2015-10-14
swagger spec fixes

nmonterroso:
@Wikia/services-team 

fixing the helios swagger spec to have correct operation ids, and combining definitions where the paths are actually identical (`POST /users/{user_name}/password` and `PUT /users/{user_id}/password`), and separating them by method in the definition.


ArturKlajnerok:
Lgtm :+1:

PR #133, opened 2015-10-15
adding resource to generate an access token

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-893

Adding `POST /users/{userId}/tokens` resource, ~~which needs to be authenticated with `THE-SCHWARTZ`~~, and will generate an access token for the specified `userId`


controllers/users_controller.go
```diff
@@ -327,3 +330,12 @@ func (controller *UsersController) forceLogoutAllSessions(w http.ResponseWriter,
 		utils.NoContent().Write(w)
 	}
 }
+
+func (controller *UsersController) forcedTokenGenerationHandler(w http.ResponseWriter, r *http.Request, params map[string]string) {
```
ArturKlajnerok: Can we create new `token_controller.go` and start to move there resources like this one, so we can deprecate other controllers at some point of breaking it apart ?

controllers/users_controller.go
```diff
@@ -327,3 +330,12 @@ func (controller *UsersController) forceLogoutAllSessions(w http.ResponseWriter,
...
+	resp := controller.server.NewResponse()
+	defer resp.Close()
+	defer osin.OutputJSON(resp, w, r)
+	userId := utils.ParseInt64(params["user_id"])
+
+	addAccessTokenToResponse(controller, userId, models.AuthenticatedNormalPass, resp, r)
```
ArturKlajnerok: Can we create json with access token directly in redis for this resource and move away from using `osin.OutputJSON` and `FinishAccessRequest`? If we add  valid `AccessData` via `SaveAccess()` the token should valid for further usage.

nmonterroso: `AccessData` is also an `osin` package struct, so we aren't really reducing our reliability on `osin` if we were to do that. Also in doing that we'd have to call `GenerateAccessToken` in the `osin` package. 

I agree we should reduce our dependency on `osin`, but we should do that as a separate effort where we have a clear plan

controllers/users_controller.go
nmonterroso: good idea. added in 0eba63b

controllers/token_controller_e2e_test.go
```diff
@@ -0,0 +1,50 @@
+package controllers
+import (
+	"github.com/stretchr/testify/suite"
+	"github.com/Wikia/go-commons/apiclient"
+	"github.com/Wikia/helios/helpers/testhelpers"
+	"github.com/Wikia/helios/helpers"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"github.com/Wikia/helios/models"
+"testing"
```
sqreek: formatter

nmonterroso: 76d5ee8

controllers/users_controller.go
sqreek: Is this should be removed?
it's duplicated code

nmonterroso: Oops, I forgot to remove it.

nmonterroso: 3268375

controllers/token_controller.go
```diff
@@ -0,0 +1,68 @@
...
+func (c *TokenController) GetServer() *osin.Server {
+	return c.osin
+}
+
+func (c *TokenController) GetRedisStorage() *storage.RedisStorage {
+	return c.storage
+}
+
+// resources
+func (c *TokenController) forcedTokenGenerationHandler(w http.ResponseWriter, r *http.Request, params map[string]string) {
```
ArturKlajnerok: can we just name it `tokenGenerationHandler` ? :smile_cat:

ArturKlajnerok:
Please merge master so test should pass, other than that LGTM :+1:

PR #136, opened 2015-10-22
updating swagger basePath property

nmonterroso:
@Wikia/services-team 

Now that helios is inaccessibly through `/helios` at the API gateway, we should update the Swagger docs to use `/auth`, so the central swagger ui can still work with the service


ArturKlajnerok:
Lgtm :+1:

PR #137, opened 2015-10-23
UserPermissionGateway

nmonterroso:
@Wikia/services-team 
https://wikia-inc.atlassian.net/browse/SERVICES-946

Adding UserPermissionGateway, which supports checking if a user is prevented from logging in with a specific method. If so will return `401`.

I also made a lot of changes to clean up the controller constructors, since there was a lot of duplicated code there, and for the controllers to use interfaces instead of direct implementations where it was easy to do so. 

I also removed the `RedisStorageForTests` stuff, since to use that for the `oauth_controller` tests we'd have to implement all the `SortedSet` functions from redis into that class. Instead, it's now changed to have a stub implementation for the `Storage` layer.


controllers/facebook_controller.go
```diff
@@ -52,29 +49,9 @@ func NewFacebookController(
...
-	userValidator := domain.NewUserValidator(validationFunctions...)
-
-	controller.userService = service.NewUserService(userStorage, userValidator,
-		wikiaClient, userPreferenceGateway, service.NewAuthService(systemContainer))
-
-	controller.facebookService = service.NewFacebookService(
-		userStorage, gateway.NewFacebookGateway(facebookConfig), controller.userService)
+	controller.permissionGateway = permissionGateway
+	controller.userService = userService
+	controller.facebookService = facebookService
```
drsnyder: Minor: Should we consolidate on the `SystemContainer` as the sole object to inject to the controllers? This is obviously a major improvement but it would be great if all of the dependencies could be injected via one object.

server/helios.go
```diff
@@ -85,35 +87,64 @@ func (helios *Helios) Run(configPath string) {
...
+	wikiaServiceGateway := gateway.NewWikiaServiceGateway(conf.WikiaServices.Tag)
+	userPreferenceGateway := gateway.NewUserPreferenceGateway(wikiaServiceGateway)
+	userPermissionGateway := gateway.NewUserPermissionGateway(wikiaServiceGateway)
+	authService := service.NewAuthService(systemContainer)
+	userService := service.NewUserService(
+		userStorage,
+		userValidator,
+		wikiaClient,
+		userPreferenceGateway,
+		authService)
```
drsnyder: This is what I was trying to avoid with the `SystemContainer`. If all the services and controllers accept the `SystemContainer` then you would never have to modify the dependent constructors again if a service is added or removed. You would only have to modify the `SystemContainer`.

gateway/user_permission.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/Wikia/go-commons/logger"
+	"github.com/Wikia/helios/errorlist"
+	"io/ioutil"
+	"net/http"
+)
+
+const (
+	UserPermissionServiceName   = "user-permissions"
+	LoginMethodUsernamePassword = "login_blocked_username_password"
+	LoginMethodFacebook         = "login_blocked_facebook"
```
drsnyder: I think I would prefer to invert the naming here and use something like `can_login_with_facebook`.

controllers/facebook_controller.go
```diff
@@ -123,6 +100,12 @@ func (controller *FacebookController) tokenHandler(w http.ResponseWriter, r *htt
 			errorlist.GenericDataAccessError)
 	} else if !errors.IsEmpty() {
 		setRequestErrors(resp, errors, "unable to lookup user from access token")
+	} else if controller.permissionGateway.LoginMethodBlocked(
```
drsnyder: I would prefer to see a layer of abstraction above the gateway. Something like:

```go
if !controller.authorizationService.authorized(userId, action) { # or authorizedTo, or authorizeAction or can
   errors.Add(gateway.LoginMethodBlocked)
   ...
}
```

controllers/facebook_controller.go
nmonterroso: I'm not really sold on having a `SystemContainer` with all the dependencies, seems odd to have every controller have a reference to everything, even if it only needs a single entry. Instead I think we should invest in sane dependency resolution/injection in Go. I've played with [`facebook's dependency injection library`](https://github.com/facebookgo/inject), and although it's not perfect it's better than what we have now.

gateway/user_permission.go
nmonterroso: I would too, but that requires having a default permission with an override per group in MW. When I tried doing that, the MW api returned the permission for all users, not just the ones that had it enabled, and it doesn't return the values so it's not really doable I think. Ex:

```php
$wgGroupPermissions['*']['login-allow-facebook'] = true;
$wgGroupPermissions['*']['login-allow-username-password'] = true;
$wgGroupPermissions['restricted-login']['login-allow-username-password'] = false;
$wgGroupPermissions['restricted-login']['login-allow-facebook'] = false;
```

Resulted in:
```
// for user not in group
"perms": [
  "login-allow-username-password",
  "login-allow-facebook",
]

// for user in group
"perms": [
  "login-allow-username-password",
  "login-allow-facebook",
]
```

controllers/facebook_controller.go
nmonterroso: Hm I think maybe it would make more sense to just rename the current gateway to `AuthorizationService` in that case, since why have an abstraction on top of another abstraction if we can avoid it?

Also I think the method should stay named `LoginMethodBlocked`. I'm specifically ignoring any errors that come up during the http request to the permission service and just returning `false` in that case, since 99% of logins won't be blocked anyway. So I think that it's a good idea to have 

```
func (s *authorizationService) LoginMethodBlocked(userId, method string) bool {
  blocked, err :=  s.authorized(userId, method)
  if err != nil {
    // log
    return false
  }

  return blocked
}

func (s *authorizationService) authorize(userId, action string) (bool, error) {
  // ask service
}
```

controllers/facebook_controller.go
drsnyder: The `SystemContainer` is effectively a "light" DI container. It's not that much different than what we are doing on the PHP side [here](https://github.com/Wikia/app/blob/dev/includes/User.php#L283-L299). The major difference is that the setup is manual. The impact on testing and telescoping constructors should be the same.

I'm ok either way but I think we should do something.

controllers/facebook_controller.go
drsnyder: > Hm I think maybe it would make more sense to just rename the current gateway to AuthorizationService in that case, since why have an abstraction on top of another abstraction if we can avoid it?
:+1:

nmonterroso: 727f65c

controllers/facebook_controller.go
nmonterroso: Yah that's a good point. Let's create a ticket in our backlog to do that? We can take this half-step here and then clean it up even more with the `SystemContainer`, since at least now the controller constructors don't manually create the objects, they jsut assign whatever comes into the constructor args and assigns it to a local copy.

drsnyder: Created https://wikia-inc.atlassian.net/browse/SERVICES-953 to track the follow up regarding DI.

PR #138, opened 2015-10-29
fixing permission names

nmonterroso:
@Wikia/services-team 
fixing the permission names to be consistent with what's in MW


ArturKlajnerok:
Lgtm :+1:

PR #139, opened 2015-10-29
changing response types in WikiaServiceGateway

nmonterroso:
@Wikia/services-team 

Changing response type in `WikiaServiceGateway` from `*http.Response` to `WikiaServiceGatewayResponse` since with the current implementation callers would be responsible for closing the body if we wanted to get it outside of the `request` function, which we do for reading from the `user-permissions` service.


gateway/user_preference.go
```diff
@@ -57,9 +56,9 @@ func (gateway *UserPreferenceGatewayValue) UpdateUserPreferences(userId int64, p
 		return false, UserPreferenceServiceError
 	}
 
-	if resp.StatusCode != 204 {
-		if resp.Body != nil {
-			bs, _ := ioutil.ReadAll(resp.Body)
+	if resp.GetStatus() != 204 {
+		if resp.GetBody() != nil {
+			bs := resp.GetBody()
```
ArturKlajnerok: 'bs' is not need `resp.GetBody()` can be used directly in log.

nmonterroso: 0196b09

ArturKlajnerok:
Lgtm :+1:

PR #140, opened 2015-11-02
Permissions auth

nmonterroso:
@Wikia/services-team @jcellary 
Adding authentication header to user-permissions service usage in Helios, since the service will soon require authentication.


ArturKlajnerok:
Lgtm :+1:

PR #141, opened 2015-11-03
[SERVICES-833] Removed application/v2+json from request to user preferences

ArturKlajnerok:
/cc @Wikia/services-team


sqreek:
tests?

ArturKlajnerok:
Integration test was failing because of lack of request compatibility with user preference service. Fixed that :+1:

nmonterroso:
LGTM

PR #144, opened 2015-11-06
[SERVICES-972] Blocking reset password for restricted-login group

ArturKlajnerok:
This introduced block on reset password resource for users that have the `password-reset-restrict` permission.

/cc @Wikia/services-team


service/authorization_test.go
```diff
@@ -45,18 +45,48 @@ func (s *AuthorizationServiceTestSuite) TestLoginMethodNotBlocked() {
...
+
+	assert.True(s.T(), s.authorizationService.ResetPasswordBlocked(AuthorizationTestUserId))
+}
+
 func (s *AuthorizationServiceTestSuite) TestPermissionCheckUnauthorized() {
 	response := createResponse(http.StatusUnauthorized, "")
 	s.services.On("GET", UserPermissionServiceName, s.authorizationService.getPermissionPath(AuthorizationTestUserId), createRequestHeaders(AuthorizationTestUserId)).Return(response, nil)
 
-	assert.False(s.T(), s.authorizationService.LoginMethodBlocked(AuthorizationTestUserId, LoginMethodUsernamePassword))
+	assert.False(s.T(), s.authorizationService.checkUserPermission(AuthorizationTestUserId, LoginMethodUsernamePassword))
```
nmonterroso: should these stay as checking `LoginMethodBlocked`? That's what external packages will call

nmonterroso:
I think we should add specific tests for the reset password functionality in the authorization service. Right now the tests pass because the underlying logic is identical to the `LoginMethodBlocked` method, but if we ever change that and it causes an error, it'll go unnoticed in the tests because the tests are testing the underlying logic, not what other libraries will call

service/authorization_test.go
ArturKlajnerok: I moved common common code around checking if user has a specyfic permission to `checkUserPermission` that's way tests around that common functionality are calling now`checkUserPermission`. And now `LoginMethodBlocked` is basically a wrapper around `checkUserPermission`.

ArturKlajnerok:
But then we will have pretty the same tests for `LoginMethodBlocked` and `ResetPasswordBlocked`.

nmonterroso:
Yah but the contract the `AuthorizationService` has with the outside world is "can this user log in using this method" and "can this user reset their password", not the private helper method that isn't guaranteed to be called by the contracted methods

nmonterroso:
Or we change `AuthorizationService` to only have one method: `IsAllowed(userId, permission string) bool`

ArturKlajnerok:
I think that using one `IsAllowed` method is cleaner, so I changed adjusted that.

nmonterroso:
LGTM

